PR #4, opened 2016-05-04
Esa 72 - Add internationalization

vguzmanp:


.tx/config
```diff
@@ -0,0 +1,11 @@
+[main]
+host = https://www.transifex.com
+
+[VAS.VAS]
```
vguzmanp: We will need to change this as soon as the transifex project is created.

aameijide:
Checked with Longomatch, the build/m4 has more files here. Maybe some files must not been pushed

vguzmanp:
Ready to review again

aameijide:
👍

ylatuya:
Missing autogen.sh script

po/LINGUAS
```diff
@@ -0,0 +1,2 @@
+# please keep this list sorted alphabetically
+#
```
ylatuya: Include the same languages as in longomatch

vguzmanp: Including the languages but not the translated .po files makes `make check` to fail. Is this what we want?

ylatuya: We should include the translations too, copied from longomatch and cleaned up. There should be a script to automate it for when we merge other libraries from longomatch

vguzmanp: Fixed using `intltool-update`

PR #5, opened 2016-05-05
Migrate Multimedia, Cairo, GUI.Helpers, some Services to VAS

aoller:
Change Cairo to VAS

partial migration

interfaces

Move partially EventsBroker to VAS

Events broker change

Move services part I

Move services part II

Move services part III

migrate code to VAS solution

Migrate some Services classes

migrate PlaylistManager Service to VAS


ylatuya:
I think Config should now be splitted in 2 parts:

1. The application state (path initialization, registry for deps injection, etc...)
2. The real application configuration, only properties that can be changed by the user which are now in ConfigState.

Also all the interfaces hardcoded in Config,  should now be provided with dependency injection using a  LongoMatch.Core.Common.Registry.

An application at startup would register the interface App.Registry.register(IDrawingToolkit, GtkDrawingToolki) and would use App.Registry.Retrieve<IDrawingToolkit>() instead of Config.IDrawingTookit.

VAS.Services/EventsManagerBase.cs
```diff
@@ -80,7 +80,7 @@ protected virtual void HandlePlayLoaded (TimelineEvent play)
 			loadedPlay = play;
 		}
 
-		protected abstract void HandlePlaylistElementSelectedEvent (Playlist playlist, IPlaylistElement element, bool playing);
+		protected abstract void HandlePlaylistElementLoaded (Playlist playlist, IPlaylistElement element);
```
aameijide: Why there's no HandleLoadPlaylistElement if there's a Handler for that?

VAS.Tests/Services/TestPlaylistManager.cs
```diff
@@ -0,0 +1,344 @@
+﻿//
+//  Copyright (C) 2015 vguzman
```
aameijide: Copyright info must be Fluendo S.A.

VAS.Tests/Services/TestPlaylistManager.cs
```diff
@@ -0,0 +1,344 @@
...
+		{
+			mockGuiToolkit = new Mock<IGUIToolkit> ();
+			Config.GUIToolkit = mockGuiToolkit.Object;
+			Config.RenderingJobsManger = mockVideoRenderer.Object; 
+
+			plmanager = new PlaylistManager ();
+			plmanager.Start ();
+			plmanager.Player = mockPlayerController.Object;
+
+			OpenProject (new VAS.Tests.Utils.ProjectDummy ());
```
aameijide: It's better to use here an alias

PR #8, opened 2016-05-20
ESA-72 & autotoolize

vguzmanp:


AssemblyInfo/AssemblyInfo.cs
```diff
@@ -1,45 +1,54 @@
...
-// and "{Major}.{Minor}.{Build}.*" will update just the revision.
+// The assembly version has following format :
+//
+// Major.Minor.Build.Revision
+//
+// You can specify all values by your own or you can build default build and revision
+// numbers with the '*' character (the default):
 
-[assembly: AssemblyVersion ("1.0.*")]
+[assembly: AssemblyVersion("0.2")]
```
vmartos: This assemblyVersion is set to 0.2 for any reason? It's changed during build time by the Longomatch release tag?

build/build_env.mk
```diff
@@ -0,0 +1,8 @@
+LM_DEPS = \
```
vmartos: Maybe should be VAS_DEPS, as we are in the VAS repo?

vmartos:
Code review Done

AssemblyInfo/AssemblyInfo.cs
vguzmanp: I set it to 0.2 as a placeholder. This way we avoid using the git-version-gen which, without tags, would make the version `UNKNOWN`

build/build_env.mk
vguzmanp: Actually it's never used in VAS, we could remove it.
At the beginning I thought of using the VAS `build` folder and symlink to it from LongoMatch, but it's probably easier to have a copy

vguzmanp:
Ready to review again

vmartos:
:+1:

PR #9, opened 2016-05-23
Add Prism nuget package and The EventsAggregator wrapper

vmartos:
Also added a static reference in Config to use it

Add EventTypes file with a NewTagEvent


vguzmanp:
:+1:

PR #12, opened 2016-05-30
Change NuGet.exe download method to curl

aoller:


ylatuya:
👍  with the 2 commits squashed (it's easier to revert this way)

PR #17, opened 2016-06-07
ESA-65

vmartos:


VAS.Drawing/CanvasObjects/Dashboard/CategoryObject.cs
```diff
@@ -667,6 +684,31 @@ void DrawBackbuffer (IDrawingToolkit tk)
 			tk.End ();
 		}
 
+		void DrawHotkey (IDrawingToolkit tk)
+		{
+			if (ShowHotkey) {
```
ylatuya: I would do if (!ShowHotkey) -> return as defined in the coding style

VAS.Drawing/CanvasObjects/Dashboard/CategoryObject.cs
```diff
@@ -442,12 +453,17 @@ void DrawHeader (IDrawingToolkit tk)
...
 					width = Width;
 					height = Height - HeaderHeight;
 					pos = new Point (Position.X, Position.Y + HeaderHeight);
 					fontSize = StyleConf.ButtonNameFontSize;
 					ellipsize = false;
+				} else {
+					width = HeaderTextWidth;
+					height = HeaderHeight;
+					pos = new Point (Position.X + 5, Position.Y + 5);
+					fontSize = StyleConf.ButtonHeaderFontSize;
```
ylatuya: Why do you need to render this differently based on wether there is an icon or not?

VAS.Drawing/CanvasObjects/Dashboard/CategoryObject.cs
```diff
@@ -315,10 +324,12 @@ public override LinkAnchorObject GetAnchor (IList<Tag> sourceTags)
 
 		public override void ClickPressed (Point p, ButtonModifier modif)
 		{
-			if (Mode == DashboardMode.Edit) {
+			if (Mode == DashboardMode.Edit || Button.ShowSettingIcon) {
 				editClicked = CheckRect (p, editRect, editbutton);
-				return;
+				if (editClicked && Button.ShowSettingIcon)
```
ylatuya: I don't understand why this second check for Button.ShowSettingIcon is needed

VAS.UI.Gtk2/UI/Menu/PlaysMenu.cs
```diff
@@ -78,36 +83,15 @@ public void ShowTimelineMenu (Project project, List<TimelineEvent> plays, EventT
...
 				string label = String.Format ("{0} ({1})", Catalog.GetString ("Delete"), plays.Count ());
 				del.SetLabel (label);
-				label = String.Format ("{0} ({1})", Catalog.GetString ("Export to video file"), plays.Count ());
-				render.SetLabel (label);
-				label = String.Format ("{0} ({1})", Catalog.GetString ("Duplicate "), plays.Count ());
-				duplicate.SetLabel (label);
 			}
 		
-			if (moveCat.Visible) {
+			/*if (moveCat.Visible) {
```
ylatuya: Commented code should be deleted

VAS.Drawing/CanvasObjects/Dashboard/CategoryObject.cs
vmartos: when there is an Icon the text is displayed in Position.Y + HeaderHeight, so it is displayed in the center of the button. But when there is no icon, the text should be displayed in Position.Y+5 so it's almost at the top of the button, based on the Story Design attachments

VAS.Drawing/CanvasObjects/Dashboard/CategoryObject.cs
vmartos: Because now we can click the edit button if we are not in DashboardMode.Edit && the ShowSettingIcon is true, and if the click is outside the editRect we need to continue and not return immediately if editClicked is false. probably it will have more sense if the if statement is:
if(editClicked || Mode == DashboardMode.Edit) -> return 
because in DashboardMode.Edit we only need to catch if editClicked is true or false and then return

ylatuya:
👍

VAS.Core/Filters/EventsFilter.cs
```diff
@@ -196,14 +196,70 @@ protected virtual void UpdateVisibleCategories ()
...
+			}
+		}
 
 		protected virtual void EmitFilterUpdated ()
 		{
 			if (!Silent && FilterUpdated != null)
 				FilterUpdated ();
 		}
+
+		bool IsVisibleByCategory (TimelineEvent play)
```
ylatuya: I would make those protected virtual so they can be overriden by the parent class.

PR #20, opened 2016-06-10
RA-108

ylatuya:


VAS.UI.Gtk2/gtk-gui/gui.stetic
```diff
@@ -162,4 +162,4 @@
       </widget>
     </child>
   </widget>
-</stetic-interface>
\ No newline at end of file
+</stetic-interface>
```
vguzmanp: In Longomatch the whole SplashScreen widget is removed from the gui.stetic, but here it isn't added. (https://github.com/LongoMatch/longomatch/pull/115/commits/a1c14852b7c938692ee7d83774d513ecb01ae066#diff-beec728c064280195715eb68fe0abca8L12317)

ylatuya: This widget is now build manually in Build () without using stetic. That because the size of the widget must adapt to the size of the image used and using stetic which calls Show() at the of Build() we couldn't change the size of the widget without seeing a glitch.

vguzmanp: It would be useful to add that to either the commit message or a comment in the code.

PR #21, opened 2016-06-14
Ra 81 - Mainly PlayerObject

vguzmanp:


VAS.Drawing.Cairo/CairoBackend.cs
```diff
@@ -28,6 +28,7 @@
 using Image = VAS.Core.Common.Image;
 using LineStyle = VAS.Core.Common.LineStyle;
 using Point = VAS.Core.Common.Point;
+using VAS.Core;
```
ylatuya: Sort using

VAS.Drawing.Cairo/CairoBackend.cs
```diff
@@ -154,7 +155,7 @@ public CairoBackend ()
 
 		public ISurface CreateSurface (string filename, bool warnOnDispose = true)
 		{
-			Image img = new Image (filename);
+			Image img = Resources.LoadImage (filename);
```
ylatuya: Image uses a full path to the image while LoadImage uses a relative path.

vguzmanp: Hm... I changed this because something was calling this method with a relative path. Let me check again

vguzmanp: I see now, but anyways Resources.LoadImage should be safe: `Path.Combine` with an absolute path returns the same absolute path.

ylatuya:
👍

PR #23, opened 2016-06-15
VAS-6

vmartos:


aoller:
👍

VAS.Services/PlayerController.cs
```diff
@@ -445,6 +445,7 @@ protected virtual bool PlaylistSeek (Time time, bool accurate = false, bool sync
 					Log.Debug (String.Format ("There is no playlist element at {0}.", time));
 					return false;
 				}
+				LoadedPlaylist.SetActive (elementAtTime);
```
ylatuya: Shouldn't this be done in LoadPlaylistEvent for every event loaded in the playlist ?

VAS.DB/DocumentsSerializer.cs
```diff
@@ -323,6 +323,9 @@ static string GetDocumentType (IStorable storable)
 		{
 			Type type = storable.GetType ();
 
+			if (DocumentTypeBaseTypes.ContainsValue (type.Name))
+				return type.Name;
```
ylatuya: This check is wrong, it should be:
if (DocumentTypeBaseTypes.ContainsKey (type))
return DocumentTypeBaseTypes[type]

vmartos:
Ready to codereview again

aoller:
👍

PR #24, opened 2016-06-16
Ra 80

ylatuya:


VAS.Core/Store/Player.cs
```diff
@@ -97,13 +97,32 @@ public void Dispose ()
...
+		/// <summary>
+		/// Gets or sets a value indicating whether this <see cref="VAS.Core.Store.Player"/> is currently tagged.
+		/// </summary>
+		/// <value><c>true</c> if tagged; otherwise, <c>false</c>.</value>
+		[JsonIgnore]
+		public bool Tagged {
+			get;
+			set;
+		}
+
```
vguzmanp: Shouldn't these be properties of the Card object? We won't need them in the DB or anything like that

PR #25, opened 2016-06-20
Add support for multiple data directories

aoller:


dfernandezramos:
👍

ylatuya:
I think we need a helper method to resolve datadir paths since we will use it in several places.
Something like GetDataFilepath(string filename) that iterates over all dirs and throws an IO exception of the file is not found

VAS.Core/App.cs
```diff
@@ -37,6 +37,11 @@ public abstract class App
 			set;
 		}
 
+		public List<string> DataDir {
```
ylatuya: This property should be below the static method with the other properties.

ylatuya:
There is already a function RelativeToPrefix (string relativePath) to resolve paths relative to the prefix, we could follow the same scheme to resolve paths relative to the data dir

VAS.Core/App.cs
```diff
@@ -55,7 +60,8 @@ public static void Init (App appInit, string evUninstalled, string softwareName,
 
 			if (Environment.GetEnvironmentVariable (evUninstalled) != null) {
 				App.Current.baseDirectory = Path.GetFullPath (".");
-				App.Current.dataDir = "../data";
+				App.Current.DataDir.Clear ();
```
ylatuya: Clear() is called in both cases. Why not leaving the DataDir unitialized, and initialize it here correctly? I prefer the DataDir being null rather to initialize to a value that's useless like "."

VAS.Core/App.cs
```diff
@@ -25,18 +26,31 @@ public abstract class App
...
 		public string baseDirectory = ".";
 		public string configDirectory = ".";
-		public string dataDir = ".";
 
 		/* State */
 		public EventsAggregator EventsAggregator;
 
 		protected StyleConf style;
 
+		public List<string> DataDir {
```
dfernandezramos: As Andoni said, this should be below the static method with the other properties :)

VAS.Core/App.cs
```diff
@@ -55,7 +69,7 @@ public static void Init (App appInit, string evUninstalled, string softwareName,
 
 			if (Environment.GetEnvironmentVariable (evUninstalled) != null) {
 				App.Current.baseDirectory = Path.GetFullPath (".");
-				App.Current.dataDir = "../data";
+				App.Current.DataDir.Add (appInit.RelativeToPrefix ("../data"));
```
dfernandezramos: this should be Current instead appInit

VAS.Core/App.cs
```diff
@@ -69,7 +83,7 @@ public static void Init (App appInit, string evUninstalled, string softwareName,
 				}
 				if (!Directory.Exists (Path.Combine (App.Current.baseDirectory, "share", softwareName)))
 					Log.Warning ("Prefix directory not found");
-				App.Current.dataDir = Path.Combine (App.Current.baseDirectory, "share", softwareName.ToLower ());
+				App.Current.DataDir.Add (appInit.RelativeToPrefix (Path.Combine ("share", softwareName.ToLower ())));
```
dfernandezramos: this should be Current instead appInit

VAS.Core/App.cs
dfernandezramos: It works with appInit too but to make it clearer, as all Init method works, we just use appInit to initialise Current. Then use Current for whatever you need :)

dfernandezramos:
👍

PR #28, opened 2016-06-27
RA-82: Add dll.config to VAS.Multimedia check for libvas

vmartos:
-Configure.ac check libvas deps and add to make


turran:
Please, create a VAS task to refactor every libvas (old libcesarplayer, old liblgm) API and internals to match the new namespace
:+1:

PR #32, opened 2016-07-07
Add HotKey related stuff to manage hotkeys dynamically

vmartos:
- Move Keyboard to VAS.Core.Hotkeys and add additional
Implementation functions
- Enable temporal Context to enable focus, etc.
- Make KeyContextManager create the keyboard. Unic instance
- Add KeyActionMappings API to the Keyboard. (Not just
in the constructor)
- Add HotKey Fallback functionality to go through
the previous Hotkey method
- ActualContext is now a list
- Contexts priorities, last added more priority
- Add a GlobalContext with the lower priority
- Add Tests

### DIFF
--- a/VAS.Core/App.cs
+++ b/VAS.Core/App.cs
@@ -4,6 +4,7 @@ using System.IO;
 using VAS.Core;
 using VAS.Core.Common;
 using VAS.Core.Events;
+using VAS.Core.Hotkeys;
 using VAS.Core.Interfaces;
 using VAS.Core.Interfaces.Drawing;
 using VAS.Core.Interfaces.GUI;
@@ -32,6 +33,12 @@ namespace VAS
 
 		/* State */
 		public EventsBroker EventsBroker;
+
+		public KeyContextManager KeyContextManager {
+			get {
+				return KeyContextManager.Instance;
+			}
+		}
 
 		public List<string> DataDir {
 			get {
--- a/VAS.Core/Hotkeys/KeyAction.cs
+++ b/VAS.Core/Hotkeys/KeyAction.cs
@@ -1,0 +1,120 @@
+﻿using System;
+using VAS.Core.Store;
+
+namespace VAS.Core.Hotkeys
+{
+	/// <summary>
+	/// Object used to bind an Action to a Hotkey and actionName
+	/// </summary>
+	public class KeyAction : IComparable, IEquatable<KeyAction>
+	{
+		/// <summary>
+		/// The GUID of the KeyAction
+		/// </summary>
+		public string guid;
+
+		/// <summary>
+		/// Gets or sets the name of the action
+		/// </summary>
+		/// <value>The name of the action.</value>
+		public string ActionName { get; set; }
+
+		/// <summary>
+		/// Gets or sets the key that performs the action
+		/// </summary>
+		/// <value>The key.</value>
+		public HotKey Key { get; set; }
+
+		/// <summary>
+		/// Gets or sets the action to perform
+		/// </summary>
+		/// <value>The action.</value>
+		public Action Action { get; set; }
+
+		/// <summary>
+		/// Sets the KeyAction Priority
+		/// Lower numbers = higher priority
+		/// </summary>
+		/// <value>The priority.</value>
+		public int Priority { get; set; }
+
+		/// <summary>
+		/// Gets or sets a value indicating whether this KeyAction continues the chain
+		/// or breaks it and other KeyAction with the same ActionName doesn't execute
+		/// </summary>
+		/// <value><c>true</c> if continue chain; otherwise, <c>false</c>.</value>
+		public bool ContinueChain { get; set; }
+
+		public KeyAction (string actionName, HotKey key, Action action, int priority = 999, bool continueChain = true)
+		{
+			this.ActionName = actionName;
+			this.Key = key;
+			this.Action = action;
+			this.Priority = priority;
+			this.ContinueChain = continueChain;
+			guid = Guid.NewGuid ().ToString ("N");
+		}
+
+		public bool Equals (KeyAction action)
+		{
+			if (action == null) {
+				return false;
+			}
+
+			return this.guid == action.guid;
+		}
+
+		#region IComparable implementation
+
+		int IComparable.CompareTo (object obj)
+		{
+			if (obj as KeyAction == null) {
+				return 1;
+			}
+			return Priority - ((KeyAction)obj).Priority;
+		}
+
+		#endregion
+
+		#region Operators
+
+		static public bool operator == (KeyAction a, KeyAction b)
+		{
+			// If both are null, or both are same instance, return true.
+			if (System.Object.ReferenceEquals (a, b)) {
+				return true;
+			}
+
+			// If one is null, but not both, return false.
+			if (((object)a == null) || ((object)b == null)) {
+				return false;
+			}
+			return a.Equals (b);
+		}
+
+		static public bool operator != (KeyAction a, KeyAction b)
+		{
+			return !(a == b);
+		}
+
+		#endregion
+
+		#region Overrides
+
+		public override bool Equals (object obj)
+		{
+			if (obj is KeyAction) {
+				KeyAction ka = obj as KeyAction;
+				return Equals (ka);
+			} else
+				return false;
+		}
+
+		public override int GetHashCode ()
+		{
+			return guid.GetHashCode ();
+		}
+
+		#endregion
+	}
+}
--- a/VAS.Core/Hotkeys/KeyContext.cs
+++ b/VAS.Core/Hotkeys/KeyContext.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace VAS.Core.Hotkeys
+{
+	/// <summary>
+	/// Object to store all KeyActions in a Context
+	/// </summary>
+	public class KeyContext
+	{
+		/// <summary>
+		/// Gets the KeyActions
+		/// </summary>
+		/// <value>The key actions.</value>
+		public List<KeyAction> KeyActions {
+			get;
+		}
+
+		public KeyContext ()
+		{
+			KeyActions = new List<KeyAction> ();
+		}
+
+		/// <summary>
+		/// Adds a KeyAction
+		/// </summary>
+		/// <param name="keyAction">Key action.</param>
+		public void AddAction (KeyAction keyAction)
+		{
+			if (!KeyActions.Contains (keyAction)) {
+				KeyActions.Add (keyAction);
+				KeyActions.Sort ();
+			}
+		}
+
+		/// <summary>
+		/// Removes a KeyAction
+		/// </summary>
+		/// <param name="keyAction">Key action.</param>
+		public void RemoveAction (KeyAction keyAction)
+		{
+			KeyActions.RemoveAll (x => x == keyAction);
+			KeyActions.Sort ();
+		}
+
+	}
+}
--- a/VAS.Core/Hotkeys/KeyContextManager.cs
+++ b/VAS.Core/Hotkeys/KeyContextManager.cs
@@ -1,0 +1,142 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Timers;
+using VAS.Core.Events;
+using VAS.Core.Store;
+
+namespace VAS.Core.Hotkeys
+{
+	/// <summary>
+	/// Manager to Register KeyContexts, global or Current State app context
+	/// Manages Actions when a Hotkey is pressed and compares it to the KeyContexts
+	/// </summary>
+	public sealed class KeyContextManager
+	{
+		private static readonly KeyContextManager instance = new KeyContextManager ();
+		List<KeyContext> currentKeyContexts;
+		KeyContext globalKeyContext;
+
+		private KeyContextManager ()
+		{
+			currentKeyContexts = new List<KeyContext> ();
+			globalKeyContext = new KeyContext ();
+			EnableGlobalContext = true;
+		}
+
+		/// <summary>
+		/// Gets the instance.
+		/// </summary>
+		/// <value>The instance.</value>
+		public static KeyContextManager Instance {
+			get {
+				return instance;
+			}
+		}
+
+		/// <summary>
+		/// Gets the global key context.
+		/// </summary>
+		/// <value>The global key context.</value>
+		public KeyContext GlobalKeyContext {
+			get {
+				return globalKeyContext;
+			}
+		}
+
+		/// <summary>
+		/// Gets or sets a value indicating whether to enable global context.
+		/// </summary>
+		/// <value><c>true</c> if enable global context; otherwise, <c>false</c>.</value>
+		public bool EnableGlobalContext {
+			get;
+			set;
+		}
+
+		/// <summary>
+		/// Gets the current key contexts.
+		/// </summary>
+		/// <value>The current key contexts.</value>
+		public List<KeyContext> CurrentKeyContexts {
+			get {
+				return currentKeyContexts;
+			}
+		}
+
+		/// <summary>
+		/// Adds the context to the List of Current Contexts
+		/// The Last Context Added the Higher the priority
+		/// </summary>
+		/// <param name="context">Context.</param>
+		public void AddContext (KeyContext context)
+		{
+			currentKeyContexts.Add (context);
+		}
+
+		/// <summary>
+		/// Removes the context from the Current Contexts
+		/// </summary>
+		/// <param name="context">Context.</param>
+		public void RemoveContext (KeyContext context)
+		{
+			currentKeyContexts.RemoveAll (ctx => ctx == context);
+		}
+
+		/// <summary>
+		/// Replaces the Current Context Lists with a new List
+		/// </summary>
+		/// <param name="contexts">Contexts.</param>
+		public void NewKeyContexts (List<KeyContext> contexts)
+		{
+			currentKeyContexts = contexts;
+		}
+
+		/// <summary>
+		/// Handles the key pressed.
+		/// </summary>
+		/// <param name="key">Key.</param>
+		public void HandleKeyPressed (HotKey key)
+		{
+			bool handled = false;
+		
+			for (int i = currentKeyContexts.Count - 1; i >= 0; i--) {
+				foreach (KeyAction ka in currentKeyContexts[i].KeyActions) {
+					if (ka.Key == key) {
+						ka.Action ();
+						handled = true;
+						if (!ka.ContinueChain) {
+							break;
+						}
+					}
+				}
+				if (handled) {
+					break;
+				}
+			}
+
+			if (!handled && EnableGlobalContext) {
+				foreach (KeyAction ka in globalKeyContext.KeyActions) {
+					if (ka.Key == key) {
+						ka.Action ();
+						handled = true;
+						if (!ka.ContinueChain) {
+							break;
+						}
+					}
+				}
+			}
+
+			if (!handled) {
+				FallbackKeyPressedEvent (key);
+			}
+		}
+
+		void FallbackKeyPressedEvent (HotKey key)
+		{
+			VAS.App.Current.EventsBroker.Publish<KeyPressedEvent> (
+				new KeyPressedEvent {
+					Key = key
+				}
+			);
+		}
+	}
+}
--- a/VAS.Tests/Core/Hotkeys/TestHotKeysContexts.cs
+++ b/VAS.Tests/Core/Hotkeys/TestHotKeysContexts.cs
@@ -1,0 +1,244 @@
+﻿using System;
+using System.Collections.Generic;
+using NUnit.Framework;
+using VAS.Core.Common;
+using VAS.Core.Events;
+using VAS.Core.Hotkeys;
+using VAS.Core.Store;
+
+namespace VAS.Tests.Core.HotKeys
+{
+	[TestFixture ()]
+	public class TestHotKeysContexts
+	{
+		const string PLAY_ACTION = "PLAY";
+		const string FORWARD_ACTION = "FORWARD";
+		const string FAST_FORWARD_ACTION = "FAST_FORWARD";
+		const string GLOBAL_ACTION = "GLOBAL";
+
+		HotKey spaceHotkey = Keyboard.ParseName ("space");
+		HotKey rightHotkey = Keyboard.ParseName ("Right");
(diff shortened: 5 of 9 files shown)


VAS.Core/App.cs
```diff
@@ -29,6 +30,13 @@ public abstract class App
 		/* State */
 		public EventsBroker EventsBroker;
 
+		public KeyContextManager KeyContextManager {
+			get {
+				return KeyContextManager.Instance;
+			}
+		}
+
+
```
vguzmanp: Extra whitespace

VAS.Core/Hotkeys/KeyAction.cs
```diff
@@ -0,0 +1,39 @@
...
+			this.Priority = priority;
+			this.ContinueChain = continueChain;
+			guid = Guid.NewGuid ().ToString ("N");
+		}
+
+		#region IComparable implementation
+
+		int IComparable.CompareTo (object obj)
+		{
+			if (obj == null)
```
vguzmanp: if you do `obj as KeyAction`, it protects against `obj` being a different type.

VAS.Core/Hotkeys/KeyAction.cs
```diff
@@ -0,0 +1,39 @@
...
+		public string ActionName { get; set; }
+
+		public Action<string> Action { get; set; }
+
+		public int Priority { get; set; }
+
+		public bool ContinueChain { get; set; }
+
+
+		public KeyAction (string actionName, Action<string> action, int priority = 999, bool continueChain = true)
```
vguzmanp: Please add comments for the properties and their meaning. Mainly `Priority`, higher number = lower priority, right?

vguzmanp: Also, extra whitespace

VAS.Core/Hotkeys/KeyContext.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace VAS.Core.Hotkeys
+{
+	public class KeyContext
+	{
+		List<string> keyActions;
+
+		public List<string> KeyActions {
```
vguzmanp: I don't think we should expose this property, we should access it only through Add/Remove methods.

VAS.Core/Hotkeys/KeyContext.cs
```diff
@@ -0,0 +1,39 @@
...
+			if (!KeyActions.Contains (keyAction)) {
+				KeyActions.Add (keyAction);
+			}
+		}
+
+		public void RemoveAction (string keyAction)
+		{
+			//KeyActions = KeyActions.Where (x => x != keyAction).ToList ();
+			//alternate code
+			KeyActions.RemoveAll (x => x == keyAction);
```
vguzmanp: extra comments

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+using System.Collections.Generic;
+using System.Timers;
+using VAS.Core.Events;
+using VAS.Core.Store;
+
+namespace VAS.Core.Hotkeys
+{
+	public class KeyContextManager
+	{
+		protected static KeyContextManager instance;
```
vguzmanp: Use the singleton implementation in the [Coding Style guide](https://fluendo.atlassian.net/wiki/display/TEAM/Coding+Style#CodingStyle-Singletonimplementation)

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+		protected List<KeyContext> actualKeyContexts;
+		protected KeyContext globalKeyContext;
```
vguzmanp: We don't need `protected` if we don't inherit from this class.
Also, inheriting from a `singleton` can be dangerous, as the child classes could define public constructors.

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+			}
+		}
+
+		public virtual KeyContext GlobalKeyContext {
+			get {
+				return globalKeyContext;
+			}
+			set {
+				if (value == null)
+					return;
```
vguzmanp: This looks ugly. Someone is doing `GlobalKeyContext = null` and nothing happens.

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+						}
+					});
+					SortKeyActionsPrioritiesInContext (actualKeyContexts [i]);
+				}
+			}
+		}
+
+		void EnableGlobalContext ()
+		{
+			CancelGlobalContext ();
```
vguzmanp: Cancel is already called when setting the Property (https://github.com/fluendo/VAS/pull/32/commits/187e4dbe950497f89f2ce5b5d8249186d7b7452f#diff-88d4bfb447664845533112c573d338caR40)

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+
+		public void RemoveContext (KeyContext context)
+		{
+			CancelContext ();
+			actualKeyContexts.RemoveAll (ctx => ctx == context);
+			EnableContext ();
+		}
+		//Replaces the entire stack to set a new KeyContext
+		// i.e. In a state transition
+		public void NewKeyContexts (List<KeyContext> contexts)
```
vguzmanp: use `///` comments to document methods properly

VAS.Core/Hotkeys/KeyContext.cs
vguzmanp: Also, it should be a auto property

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+					if (globalKeyContext != null) {
+						CancelGlobalContext ();
+					}
+					globalKeyContext = value;
+					EnableGlobalContext ();
+				}
+			}
+		}
+
+		public virtual List<KeyContext> ActualKeyContexts {
```
vguzmanp: `Actual` should be `Current`

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+			get {
+				return actualKeyContexts;
+			}
+		}
+
+		Dictionary<KeyContext,Dictionary<string, List<KeyAction>>>  KeyActionsByActionInContexts { get; set; }
+
+		Dictionary<string, List<KeyAction>>  KeyActionsByActionInGlobalContext { get; set; }
```
vguzmanp: Can we find a better name for these two? It's hard to read

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+
+		protected string CurrentContextRunId { get; set; }
+
+		protected string GlobalContextRunId { get; set; }
```
vguzmanp: Do we use these two anywhere?

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+				}
+			} else {
+				Dictionary<string, KeyAction> lKeys = new Dictionary<string, KeyAction> ();
+				lKeys.Add (keyAction.guid, keyAction);
+				KeyActions.Add (keyAction.ActionName, lKeys);
+			}
+
+			//Add KeyAction in KeyContexts List
+			if (actualKeyContexts != null) {
+				for (int i = actualKeyContexts.Count - 1; i >= 0; i--) {
```
vguzmanp: Why iterating backwards? a foreach will have the same result

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+			if (globalKeyContext != null) {
+				if (globalKeyContext.KeyActions.Contains (keyAction.ActionName)) {
+					KeyActionsByActionInGlobalContext [keyAction.ActionName].RemoveAll (x => x == keyAction);
+					if (KeyActionsByActionInGlobalContext [keyAction.ActionName].Count == 0)
+						KeyActionsByActionInGlobalContext.Remove (keyAction.ActionName);
+				}
+			}
+		}
+
+		public bool HandleKeyPressed (List<string> keyActions)
```
vguzmanp: the list contains action names, not keyActions

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+		{
+			bool handled = false;
+			//Iterate with the list of keyContexts since List guarantees insertion order
+			for (int i = actualKeyContexts.Count - 1; i >= 0; i--) {
+				if (KeyActionsByActionInContexts.ContainsKey (actualKeyContexts [i])) {
+					keyActions.ForEach ((keyAction) => {
+						if (KeyActionsByActionInContexts [actualKeyContexts [i]].ContainsKey (keyAction)) {
+							foreach (KeyAction ka in KeyActionsByActionInContexts[actualKeyContexts[i]] [keyAction]) {
+								handled = true;
+								ka.Action (keyAction);
```
vguzmanp: Why does the Action need the ActionName as a parameter?

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+								if (!ka.ContinueChain) {
+									break;
+								}
+							}
+						}
+					});
+				}
+				//If one KeyContext Handled the action just break
+				if (handled)
```
vguzmanp: We should add braces even in one line ifs (https://fluendo.atlassian.net/wiki/display/TEAM/Coding+Style#CodingStyle-Useofbraces{})

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+					break;
+			}
+			//ActualKeyContexts List couldn't handle action
+			//Let's see if GlobalContext Can
+			if (!handled) {
```
vguzmanp: Redundant if

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+					}
+				});
+			}
+
+			return handled;
+		}
+
+		public void HandleKeyPressed (HotKey key)
+		{
+			bool HandledWithNewManager = false;
```
vguzmanp: What is "new" manager?

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+			if (Keyboard != null) {
+				HandledWithNewManager = Keyboard.HandleKeyPressed (key);
+			}
+			if (!HandledWithNewManager) {
+				FallbackKeyPressedEvent (key);
+			}
+		}
+
+		protected KeyContextManager ()
```
vguzmanp: The constructor shouldn't be this far on the file (https://fluendo.atlassian.net/wiki/display/TEAM/Coding+Style#CodingStyle-Orderofconstructors,desctructors,properties,methodsandvariables)

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+		{
+			actualKeyContexts = new List<KeyContext> ();
+			globalKeyContext = new KeyContext ();
+			KeyActions = new Dictionary<string, Dictionary<string, KeyAction>> ();
+			KeyActionsByActionInContexts = new Dictionary<KeyContext,Dictionary<string, List<KeyAction>>> ();
+			KeyActionsByActionInGlobalContext = new Dictionary<string, List<KeyAction>> ();
```
vguzmanp: I **really** think this should be done differently.

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+		void CancelGlobalContext ()
+		{
+			KeyActionsByActionInGlobalContext = new Dictionary<string, List<KeyAction>> ();
+			GlobalContextRunId = "";
+		}
+
+		void EnableContext ()
+		{
+			CancelContext ();		
+			//1.- key list valid in actual keycontext, add by keycode to dictionary
```
vguzmanp: What does this mean?

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,313 @@
...
+		}
+
+		void SortKeyActionsPrioritiesInContext (KeyContext keyContext)
+		{
+			foreach (List<KeyAction> lk in KeyActionsByActionInContexts[keyContext].Values) {
+				lk.Sort ();
+			}
+		}
+
+		void AddKeyActionInActualContext (KeyContext ctx, KeyAction kaction)
```
vguzmanp: `Actual` --> `Current`

VAS.Core/Hotkeys/KeyContext.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace VAS.Core.Hotkeys
+{
+	public class KeyContext
+	{
+		List<string> keyActions;
```
vguzmanp: Shouldn't this be a List of _real_ `KeyAction`s? this way we avoid the `KeyActionsByActionInContexts` thing...

VAS.Core/Hotkeys/Keyboard.cs
```diff
@@ -0,0 +1,146 @@
...
+	public class KeyActionMap
+	{
+		public HotKey Key { get; set; }
+
+		public KeyAction KeyAction { get; set; }
+	}
+
+	public class Keyboard
+	{
+		protected Dictionary<HotKey, List<KeyAction>> KeyActionMapping { get; set; }
```
vguzmanp: Shouldnt this Dictionary be a `KeyActionMap`?

VAS.Core/Hotkeys/Keyboard.cs
```diff
@@ -0,0 +1,146 @@
...
+		public Keyboard (KeyActionMap[] keyActionMaps) : this (new List<KeyActionMap> (keyActionMaps))
+		{
+		}
+
+		/* Keyboard Static Helper Functions*/
+
+		public static uint KeyvalFromName (string name)
+		{
+			#if ! OSTYPE_ANDROID && ! OSTYPE_IOS
+			return Gdk.Keyval.FromName (name);
```
vguzmanp: `Gdk` is not included in this file (and it shouldn't be)

VAS.Core/Interfaces/IHotKey.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using VAS.Core.Hotkeys;
+using System.Collections.Generic;
+
+
+namespace VAS.Core
+{
+	public interface IHotKey
```
vguzmanp: Is this interface used anywhere?

VAS.Tests/Core/Hotkeys/TestHotKeysContexts.cs
```diff
@@ -0,0 +1,312 @@
...
+			App.Current.KeyContextManager.KeyActions.Clear ();
+			App.Current.KeyContextManager.GlobalKeyContext = new KeyContext ();
+			App.Current.KeyContextManager.NewKeyContexts (new List<KeyContext> ());
+		}
+
+		[Test ()]
+		public void TestKeyContextWithAllActions ()
+		{
+			RefreshCounts ();
+			RemoveAllKeyActionsAndContexts ();
```
vguzmanp: This should be in the `SetUp` method

VAS.Core/Hotkeys/KeyContextManager.cs
vmartos: Then, anyone could set to null and lose the global context, the idea on that was that as it's global it should be at least initialized at the beginning and one could add keyContexts to the global as necessary (i.e. a plugin), but never should be set to null.

VAS.Core/Hotkeys/KeyContextManager.cs
vmartos: No I Will delete them

VAS.Core/Hotkeys/KeyContextManager.cs
vmartos: foreach iterates from the first element of the list to the last, and is normally used when the iteration doesn't rely on order. On ordered iteration a for loop is more recommended and also we want first the last element added to the KeyContext Lists since it has more priority.

VAS.Core/Hotkeys/KeyContextManager.cs
vmartos: It's not a redundant if, we just try to execute an action in the global context only if Current Context couldn't handle the action first.

vguzmanp: but 3 lines before we have a `if(handled) break`,  so if it's already handled, it will never reach this if

vmartos: but this if is outside the for loop, the break is to exit the for loop of CurrentContext so the other KeyContexts in the Current Context list don't handle the action

vguzmanp: ups, true, you're right 👍

VAS.Core/Hotkeys/KeyContextManager.cs
vmartos: The comment was outdated and i removed the comment. Basically the enable context just creates a new KeyActionsInCurrentContexts with a KeyContext and a Dictionary with a ActionNames and it's related KeyActions

VAS.Core/Interfaces/IHotKey.cs
vmartos: Not now, but it should be used by all controllers, view-models and or views/panels that has HotKeys in it. To be able that the stateController get all KeyActionMaps and create a List of Contexts, add the KeyActions and KeyContexts to the KeyContextManager and Finally could add the KeyActionMaps to the Keyboard

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,135 @@
...
+		public static KeyContextManager Instance {
+			get {
+				return instance;
+			}
+		}
+
+		/// <summary>
+		/// Gets or sets the global key context.
+		/// </summary>
+		/// <value>The global key context. cann null the GlobalKeyContext remains the same</value>
```
vguzmanp: Typo `cann null...`

VAS.Core/Hotkeys/KeyContextManager.cs
vguzmanp: I'm torn here between leaving it like this, with the comment saying that null does nothing, or throwing a `NullArgumentException` or something like that...
I'm not sure which one would be better.

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,135 @@
...
+			currentKeyContexts = contexts;
+		}
+
+		/// <summary>
+		/// Handles the key pressed.
+		/// </summary>
+		/// <param name="key">Key.</param>
+		public void HandleKeyPressed (HotKey key)
+		{
+			bool Handled = false;
```
vguzmanp: local variables should be lowercase

VAS.Core/Hotkeys/KeyContextManager.cs
vguzmanp: Actually... is there any case where we want the `GlobalKeyContext` assigned? I understand we could need to Add/Remove to it, but replacing it entirely sounds dangerous.
We could make this property setter private.

VAS.Core/Hotkeys/KeyContextManager.cs
```diff
@@ -0,0 +1,135 @@
...
+		
+			for (int i = currentKeyContexts.Count - 1; i >= 0; i--) {
+				foreach (KeyAction ka in currentKeyContexts[i].KeyActions) {
```
vguzmanp: I find cleaner to have the Context handle the key, and return a bool telling whether to continue or break, but this is more of a preference, I can :+1: with this implementation

VAS.Core/Interfaces/IHotKey.cs
vguzmanp: Ok, I think I follow you here.
But why wouldn't it be a proper `KeyContext`?
They contain mostly the same information.

vmartos: The idea on that is that every widget, etc has his own "Hotkeys" (KeyActions configured), then in a transition of panels can be that two widgets and or panels contains the same priority so the List of KeyActions should be listed in the same KeyContext, That's why it should return a list of KeyActions instead of a KeyContext. basically,  A higher level should be the responsible on creating the KeyContext instead the lower level panels.

vguzmanp: Yes, I didn't mean to "register" from here, but both types contain the same information.
I was thinking on the lines of:

1. Higher level asks children for their `KeyContext`
2. Children return their `KeyContext`.
3. Higher level aggregates the different `KeyContexts` into a unique `KeyContext` iterating over the KeyActions defined there, and registers it to the Manager

With IHotkey the process would be the same, except using a different datatype.

VAS.Core/Interfaces/IHotKey.cs
```diff
@@ -1,13 +1,12 @@
...
-using System.Collections.Generic;
 
 
-namespace VAS.Core
+namespace VAS.Core.Interfaces
 {
 	public interface IHotKey
 	{
-		List<KeyAction> GetKeyAction ();
+		KeyContext GetKeyAction ();
```
vguzmanp: Hmm I didn't think of this like this...
This means that, for example, the `DashboardView` will implement `IHotkey`? Then the name is a bit confusing, it gives the impression that what implements it will **be** a hotkey, not just contain it.

vmartos: Ups I didnt changed the names

vguzmanp:
Squash everything and :+1:

PR #33, opened 2016-07-13
RA-88

dfernandezramos:


VAS.Core/Common/StateController.cs
```diff
@@ -0,0 +1,67 @@
...
+		public Task<bool> MoveTo (string transition)
+		{
+			if (current != null) {
+				if (!current.PostTransition ()) {
+					return Task.Factory.StartNew (() => false);
+				}
+			}
+			Console.WriteLine ("Moving to " + transition);
+
+			//IScreenState panel = (IScreenState)Activator.CreateInstance (destination [transition]);
```
vguzmanp: extra comments

VAS.Core/Common/StateController.cs
```diff
@@ -0,0 +1,67 @@
...
+			IScreenState panel = destination [transition] ();
+			bool ok = panel.PreTransition ();
+			if (ok) {
+				App.Current.GUIToolkit.LoadPanel (panel.Panel);// mainWindow.SetPanel (panel.Panel);
```
vguzmanp: extra comments

VAS.Core/Common/StateController.cs
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		public void Register (string transition, Func<IScreenState> panel)
+		{
+			destination [transition] = panel;
+		}
+
+		public void UnRegister (string transition)
+		{
+			// It should recover the previous one - a stack?
```
vguzmanp: Are we going to do this?
This means:

1. We have a `A` transition registered.
2. Other tool comes and registers `A` with another panel.
3. The other tool is disabled, so it unregisters `A`
4. Now `A` doesn't work.

VAS.Core/Common/StateController.cs
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		public void pushState (string state)
+		{
+			stateStack.Add (state);
+		}
+
+		public void popState (string state)
+		{
+			stateStack.Remove (state);
```
vguzmanp: This should only remove the last state that matches `state`

VAS.Core/Events/Events.cs
```diff
@@ -139,12 +139,12 @@ public class DatabaseCreatedEvent : Event
...
-		//public object Sender { get; set; }
-
-		public HotKey Key { get; set; }
-	}
+	//	public class KeyPressedEvent : Event
+	//	{
+	//		//public object Sender { get; set; }
+	//
+	//		public HotKey Key { get; set; }
+	//	}
```
vguzmanp: Commented code

VAS.Core/Interfaces/GUI/IMainController.cs
```diff
@@ -17,11 +17,15 @@
 //
 using System;
 using VAS.Core.Store;
+using Gtk;
```
vguzmanp: The interface shouldn't be using Gtk

VAS.Core/Interfaces/GUI/IMainController.cs
```diff
@@ -17,11 +17,15 @@
...
 
 namespace VAS.Core.Interfaces.GUI
 {
 	public interface IMainController
 	{
+		[Obsolete]
 		void AddExportEntry (string name, string shortName, Action<Project, IGUIToolkit> exportAction);
```
vguzmanp: Why mark as obsolete instead of remove it?

VAS.Core/Interfaces/GUI/ITool.cs
```diff
@@ -28,27 +29,37 @@ public interface ITool
...
 		/// Gets the menubar accelerator. If null, the menubar entry will not have an accelerator.
 		/// </summary>
 		/// <value>The menubar label.</value>
+		[Obsolete]
 		string MenubarAccelerator { get; }
 
 		/// <summary>
 		/// Bring up the Tool user interface using provided toolkit.
 		/// </summary>
+		[Obsolete]
```
vguzmanp: Obsolete markers

VAS.Core/Common/StateController.cs
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		public void emptyStateStack ()
```
vguzmanp: Method names should be PascalCase

VAS.Core/Interfaces/GUI/IMainController.cs
aameijide: This property is still used by LongoMatch in LongoMatch.Addins > AddinsManager.cs. As the variable is declared with the interface, we cannot erase from there.

vguzmanp: So I understand that when we refactor LongoMatch we will remove this from the interface, right? If so, add a comment saying so.

VAS.Core/Common/StateController.cs
```diff
@@ -10,11 +11,19 @@ public class StateController
 		Dictionary<string, Func<IScreenState>> destination;
 		IScreenState current;
 		List<string> stateStack;
+		List<TransitionStruct> transitionsStack;
+
+		struct TransitionStruct
+		{
+			public string key;
+			public Func<IScreenState> value;
+		}
```
vguzmanp: Why do we need this struct? Wouldn't it be enough to store the `key`?

vguzmanp: Nevermind

VAS.Core/Common/StateController.cs
```diff
@@ -39,12 +48,24 @@ public Task<bool> MoveTo (string transition)
 		public void Register (string transition, Func<IScreenState> panel)
 		{
 			destination [transition] = panel;
+			var a = new TransitionStruct { key = transition, value = panel };
+			transitionsStack.Add (a);
```
vguzmanp: The `transitionStack` is only used to recover the old transitions registered, so it could only store those that get replaced:
```
if destination.Keys.Contains(transition){
    stack.Add(...)
}
destination[transition] = panel;
```

VAS.Core/Common/StateController.cs
```diff
@@ -0,0 +1,89 @@
...
+				PushState (transition);
+			}
+			return Task.Factory.StartNew (() => ok);
+		}
+
+		public void Register (string transition, Func<IScreenState> panel)
+		{
+			destination [transition] = panel;
+			var a = new TransitionStruct { key = transition, value = panel };
+			transitionsStack.Add (a);
```
vguzmanp: Copying this here because Github shows it as outdated:https://github.com/fluendo/VAS/pull/33#discussion_r71340399

VAS.Core/Common/StateController.cs
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		public Task<bool> MoveTo (string transition)
+		{
+			if (current != null) {
+				if (!current.PostTransition ()) {
+					return Task.Factory.StartNew (() => false);
+				}
+			}
+			Console.WriteLine ("Moving to " + transition);
```
vguzmanp: This should be a Log.Debug call

VAS.Core/Common/StateController.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Linq;
+
+namespace VAS.Core
+{
+	public class StateController
+	{
+		Dictionary<string, Func<IScreenState>> destination;
+		IScreenState current;
+		List<string> stateStack;
+		List<TransitionStruct> transitionsStack;
```
vguzmanp: Can we find a better name for `transitionsStack`? One that implies that the transitions in here are the ones overwritten by another transition

VAS.Core/Interfaces/GUI/IScreenState.cs
```diff
@@ -0,0 +1,19 @@
...
+
+		bool PreTransition ();
+
+		bool PostTransition ();
+	}
+
+	public class ShortcutContext
+	{
+		// This means: whatever the context is with the Hotkeys
+	}
```
vguzmanp: This now is a `KeyContext`. We can remove the `ShortcutContext` and use the proper class

VAS.Core/Interfaces/GUI/IMainController.cs
aameijide: That's it! :-D
I will add a FIXME comment.

VAS.Core/Common/StateController.cs
aameijide: Changed to `overwrittenTransitions`. Also `TransitionStruct` has been changed to `Transition`

VAS.Core/Interfaces/GUI/IMainController.cs
```diff
@@ -22,6 +22,7 @@ namespace VAS.Core.Interfaces.GUI
 {
 	public interface IMainController
 	{
+		// FIXME: This has to desapear when Longomatch is adapted to use ITools.
```
vguzmanp: `desapear` -> `disappear`

VAS.Core/Common/StateController.cs
```diff
@@ -47,24 +48,23 @@ public Task<bool> MoveTo (string transition)
 
 		public void Register (string transition, Func<IScreenState> panel)
 		{
+			if (destination.Keys.Contains (transition)) {
+				overwrittenTransitions.Add (new Transition { key = transition, value = panel });
```
vguzmanp: `value` here should be `destination[transition]`

VAS.Core/Interfaces/GUI/IWelcomeButton.cs
```diff
@@ -23,7 +23,9 @@ public interface IWelcomeButton
 	{
 		string Title { get; set; }
 
-		Image Icon{ get; set; }
+		Image Icon { get; set; }
+
+		string TransitionName { get; set; }
```
vguzmanp: I don't think this should be like this. A welcome button doesn't need to call a transition, it could do other things - that's why the `Activate` method is there, it could do anything.

Inside the presentation button there will be a transition, but it doesn't need to be in the `IWelcomeButton` interface.

vguzmanp:
:+1:

VAS.Core/Events/EventTypes.cs
```diff
@@ -46,4 +46,11 @@ public class NewTagEvent : Event
 
 		public DashboardButton Button { get; set; }
 	}
+
+	public class KeyPressedEvent : Event
+	{
+		public HotKey Key;
+
+		public Gdk.ModifierType State;
```
ylatuya: Unused property already in Key.Modifier

VAS.Core/Interfaces/GUI/IWelcomeButton.cs
```diff
@@ -0,0 +1,30 @@
...
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using Image = VAS.Core.Common.Image;
+
+namespace VAS.Core.Interfaces.GUI
+{
+	public interface IWelcomeButton
+	{
+		string Title { get; set; }
```
ylatuya: Missing documentation

VAS.Core/Interfaces/GUI/IWelcomeButton.cs
```diff
@@ -0,0 +1,30 @@
...
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using Image = VAS.Core.Common.Image;
```
ylatuya: Use using VAS.Core.Common directly

VAS.Core/Store/Project.cs
```diff
@@ -185,6 +187,18 @@ public void Dispose ()
 			}
 		}
 
+		/// <summary>
+		/// Date of project last modification
+		/// </summary>
+		public virtual DateTime LastModified {
+			get {
+				return lastModified.ToLocalTime ();
```
ylatuya: Getter should not convert

VAS.Core/Store/Project.cs
```diff
@@ -185,6 +187,18 @@ public void Dispose ()
 			}
 		}
 
+		/// <summary>
+		/// Date of project last modification
+		/// </summary>
+		public virtual DateTime LastModified {
```
ylatuya: Use PropertyPreload

VAS.Core/Common/StateController.cs
```diff
@@ -0,0 +1,89 @@
...
+				if (!current.PostTransition ()) {
+					return Task.Factory.StartNew (() => false);
+				}
+			}
+			Log.Debug ("Moving to " + transition);
+
+			IScreenState panel = destination [transition] ();
+			bool ok = panel.PreTransition ();
+			if (ok) {
+				App.Current.GUIToolkit.LoadPanel (panel.Panel);
```
ylatuya: Async API but uses synchronous LoadPanel. As discussed before async platform specific INavigation should be used instead

VAS.Core/Common/StateController.cs
```diff
@@ -0,0 +1,89 @@
...
+		public Task<bool> MoveTo (string transition)
+		{
+			if (current != null) {
+				if (!current.PostTransition ()) {
+					return Task.Factory.StartNew (() => false);
+				}
+			}
+			Log.Debug ("Moving to " + transition);
+
+			IScreenState panel = destination [transition] ();
```
ylatuya: Does not check for errors for missing transitions

ylatuya:
Missing unit tests

VAS.Addins/ExtensionPoints/IPlugin.cs
```diff
@@ -0,0 +1,28 @@
...
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using Mono.Addins;
+
+namespace VAS.Addins.ExtensionPoints
+{
+	public interface IPlugin
+	{
+		string Name { get; }
```
ylatuya: Missing documentation

VAS.Addins/ExtensionPoints/IPlugin.cs
```diff
@@ -0,0 +1,28 @@
...
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using Mono.Addins;
```
ylatuya: Unused using

VAS.Core/Common/StateController.cs
```diff
@@ -0,0 +1,89 @@
...
+using VAS.Core.Interfaces.GUI;
+
+namespace VAS.Core
+{
+	public class StateController
+	{
+		Dictionary<string, Func<IScreenState>> destination;
+		IScreenState current;
+		List<string> stateStack;
+		List<Transition> overwrittenTransitions;
```
ylatuya: Why use an extra list here when destination could be Dictionary<string, List<Func<IScreenState>>> ?

VAS.Core/Interfaces/GUI/IScreenState.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace VAS.Core.Interfaces.GUI
+{
+	public interface IScreenState
+	{
+		IPanel Panel { get; set; }
+
+		KeyContext PanelKeyContext { get; set; }
+
+		bool PreTransition ();
```
ylatuya: Must be async

VAS.Core/Interfaces/GUI/IScreenState.cs
```diff
@@ -0,0 +1,16 @@
...
+
+		bool PostTransition ();
```
ylatuya: Must be async

VAS.Core/Common/StateController.cs
```diff
@@ -0,0 +1,89 @@
...
+		{
+			destination = new Dictionary<string, Func<IScreenState>> ();
+			stateStack = new List<string> ();
+			overwrittenTransitions = new List<Transition> ();
+		}
+
+		public Task<bool> MoveTo (string transition)
+		{
+			if (current != null) {
+				if (!current.PostTransition ()) {
```
ylatuya: Must be awaited

VAS.Core/Store/Project.cs
aameijide: A DateTime must be showed/serializated in UTC? I think this is not the expected case. We have the unit test VAS.Tests.Core.Store.TestProject.TestSerialization that is expecting local time instead of UTC and this test fails if we don't add the .ToLocalTime()

VAS.Core/Common/StateController.cs
aameijide: The same transition can be overwritten two or more times and we must be able to recover not only the last transition but also the previous from this one.

ylatuya: My comment was nos escaped correctly:

You could use:
`Dictionary<string, Stack<Func<IScreenState>>>` And have a stack with all the transitions of the same name and just pop/push

VAS.Core/Store/Project.cs
ylatuya: The setter must convert to UTC because of a serialization bug of Json.Net in Windows. But the getter does not need to convert to local time. See for example LastModified in Presentation.cs

ylatuya: It's even possible that this hack is no longuer neede in recetn versions of Json.NET and we can use a default get; set;

VAS.Core/Common/StateController.cs
aameijide: Ok, it will be clearer and comfortable as you say.

PR #42, opened 2016-07-27
Ra 88bis

aameijide:


VAS.Core/Common/StateController.cs
```diff
@@ -61,10 +59,8 @@ public void UnRegister (string transition)
 			destination.Remove (transition);
 
 			// Recover the previous one if exist, and remove from transitionsStack
-			if (overwrittenTransitions.Any (x => x.key == transition)) {
-				Transition last = overwrittenTransitions.FindLast (x => x.key == transition);
-				destination [transition] = last.value;
-				overwrittenTransitions.Remove (last);
+			if (overwrittenTransitions.ContainsKey (transition) && overwrittenTransitions [transition].Any ()) {
+				overwrittenTransitions [transition].Pop ();
```
vguzmanp: It should assign it to `destination[transition]`.

VAS.Tests/Setup.cs
```diff
@@ -48,6 +62,252 @@ public class ConfigDummy : Config
 		//Dummy class for VAS.Config
 	}
 
+	public class GUIToolkitDummy : IGUIToolkit
```
vguzmanp: This should be done with a `Mock<IGUIToolkit>`

VAS.Core/Common/StateController.cs
```diff
@@ -12,60 +12,58 @@ public class StateController
...
+				if (!postTransition) {
+					return await Task.Factory.StartNew (() => false);
 				}
 			}
 			Log.Debug ("Moving to " + transition);
 
-			IScreenState panel = destination [transition] ();
-			bool ok = panel.PreTransition ();
+			IScreenState panel = destination.ContainsKey (transition) ? destination [transition] () : null;
+			bool ok = panel != null && await panel.PreTransition ();
```
vguzmanp: Instead of returning null, it should Log an error and return false as soon as possible.

VAS.Core/Common/StateController.cs
```diff
@@ -26,19 +26,31 @@ public StateController ()
...
+
 			Log.Debug ("Moving to " + transition);
 
-			IScreenState panel = destination.ContainsKey (transition) ? destination [transition] () : null;
-			bool ok = panel != null && await panel.PreTransition ();
-			if (ok) {
-				App.Current.GUIToolkit.LoadPanel (panel.Panel);
-				current = panel;
-				PushState (transition);
+			if (destination.ContainsKey (transition)) {
```
vguzmanp: Better do this as a [guard clause](https://fluendo.atlassian.net/wiki/display/Team/Coding+Style#CodingStyle-Guardclauses).
Mainly because if the transition isn't there, we don't want the postTransition to execute.

vguzmanp:
:+1:

PR #46, opened 2016-08-10
RA-63

aoller:


VAS.Services/PlayerController.cs
```diff
@@ -575,6 +575,30 @@ public virtual void FramerateLower ()
...
+		{
+			Time jumpt = new Time { TotalSeconds = -10 };
+			Seek (CurrentTime + jumpt);
+		}
+
+		public virtual void SeekRightLong ()
+		{
+			Time jumpt = new Time { TotalSeconds = 10 };
+			Seek (CurrentTime + jumpt);
+		}
```
vguzmanp: These methods shouldn't be public, and should be refactored in a more generic method or even a lambda.

vguzmanp: Also, the values should be written in constants / configuration variables

VAS.Services/PlayerController.cs
```diff
@@ -352,7 +352,7 @@ public virtual void TogglePlay ()
 				Play ();
 		}
 
-		public virtual bool Seek (Time time, bool accurate, bool synchronous = false, bool throttled = false)
+		public virtual bool Seek (Time time, bool accurate = false, bool synchronous = false, bool throttled = false)
```
vguzmanp: Why do we need the default here?

aoller: Because in derived class it's overrided in this way.

vguzmanp:
👍 Remember to squash the last two commits

PR #51, opened 2016-08-26
Fix keyboard test

aameijide:


VAS.Tests/Setup.cs
```diff
@@ -58,6 +59,11 @@ public class ConfigDummy : Config
 		//Dummy class for VAS.Config
 	}
 
+	public class KeyboardDummy : Keyboard
```
vguzmanp: Shouldn't this implement `IKeyboard` instead of inheriting `Keyboard`?

aameijide: With that approach, we should implement our KeyboardDummy class, so it's better to derivate from Keyboard.

But, like Keyboard and KeyboardDummy are the same, it should be possible to instantiate Keyboard, but I prefer to follow the same pattern done with Config.

vguzmanp: I don't really see the point of having empty classes that aren't actually Dummy, as they keep all the behaviour of their parents, but :+1:

vguzmanp:
:+1:

PR #52, opened 2016-08-29
Ra 89 - fixes

aameijide:
* Add Fody for Oberservable Collection
* Delete App.Registry
* Fix test

### DIFF
--- a/VAS.Core/Common/StateController.cs
+++ b/VAS.Core/Common/StateController.cs
@@ -102,15 +102,8 @@ namespace VAS.Core
 				return false;
 			}
 
-			IScreenState current;
-			if ((current = LastModalState ()) != null ||
-			    (current = LastNavigationState ()) != null) {
-				bool postTransition = await current.PostTransition ();
-				if (!postTransition) {
-					Log.Debug ("Moving failed because panel " + current.Panel.PanelName + " cannot move.");
-					return false;
-				}
-			}
+			IScreenState current = LastNavigationState ();
+
 			Log.Debug ("Moving to " + transition + " in modal mode");
 			IScreenState state = destination [transition] ();
 			bool ok = await state.PreTransition (data);
@@ -213,6 +206,9 @@ namespace VAS.Core
 		/// <param name="transition">Transition.</param>
 		public bool UnRegister (string transition)
 		{
+			// Delete all existing instances
+			navigationStateStack.FirstOrDefault (x => x.Item1 == transition)?.Item2.Panel.Dispose ();
+
 			// Remove from transitions
 			bool ok = destination.Remove (transition);
 
@@ -243,6 +239,7 @@ namespace VAS.Core
 
 		async Task PopNavigationState ()
 		{
+			navigationStateStack [navigationStateStack.Count - 1].Item2.Panel.Dispose ();
 			navigationStateStack.RemoveAt (navigationStateStack.Count - 1);
 			IScreenState lastState = LastNavigationState ();
 			if (lastState != null) {
@@ -257,6 +254,7 @@ namespace VAS.Core
 				return;
 			}
 			for (int i = navigationStateStack.Count - 1; i > index; i--) {
+				navigationStateStack [i].Item2.Panel.Dispose ();
 				navigationStateStack.RemoveAt (i);
 			}
 			await App.Current.Navigation.LoadNavigationPanel (state.Panel);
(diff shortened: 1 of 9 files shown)


Weavers/Forwarder.cs
```diff
@@ -110,10 +121,24 @@ bool IsBindableBaseType (TypeDefinition type)
...
+			}
+		}
+
 		void ProcessType (TypeDefinition type)
 		{
 			LogInfo ("\t" + type.FullName);
 			foreach (var property in type.Properties) {
+
 				if (!bindableTypes.Contains (property.PropertyType.Resolve ()) &&
 				    !property.PropertyType.Name.StartsWith ("ObservableCollection")) {
```
ylatuya: This should check in observableTypes.

VAS.Core/Store/MediaFileSet.cs
```diff
@@ -28,7 +29,7 @@ namespace VAS.Core.Store
 	[Serializable]
 	[PropertyChanged.ImplementPropertyChanged]
 	[JsonObject]
-	public class MediaFileSet : List<MediaFile>, IStorable
+	public class MediaFileSet : ObservableCollection<MediaFile>, IStorable
```
ylatuya: Add aunit test to the check projects emit the event when mediafileset is changed

Weavers/Forwarder.cs
aameijide: observableTypes are added to bindableTypes in linea 103, so bindableTypes now has BindableBase and ObservableCollection types, it means, it checks both.

Weavers/Forwarder.cs
```diff
@@ -91,7 +96,13 @@ public void Execute ()
 			bindableTypes = new HashSet<TypeDefinition> (
 				ModuleDefinition.GetTypes ().Where (x => x.IsClass && IsBindableBaseType (x)));
 
-			// Process BindableBase types to inject the ConnectChild () function
+			// Get a list of all the classes deriving from ObservableCollection
+			var observableTypes = new HashSet<TypeDefinition> (
+				                      ModuleDefinition.GetTypes ().Where (x => x.IsClass && IsObservableCollectionType (x)));
+
+			observableTypes.ToList ().ForEach (x => bindableTypes.Add (x));
```
ylatuya: ToList() shouldn't be needed here

aameijide: ToList() was added in order to use linq. I change it for the normal foreach:
			foreach (var x in observableTypes) {
				bindableTypes.Add (x);
			}

Weavers/Forwarder.cs
ylatuya: So this check is no longer needed  !property.PropertyType.Name.StartsWith ("ObservableCollection")

aameijide: We shouldn't. If we delete this ObservableCollection (not inherited classes) are not setted

aameijide:
Ready to review again

VAS.Core/Common/StateController.cs
```diff
@@ -213,6 +213,9 @@ public void Register (string transition, Func<IScreenState> panel)
 		/// <param name="transition">Transition.</param>
 		public bool UnRegister (string transition)
 		{
+			// Delete all existing instances
+			navigationStateStack.Where (x => x.Item1 == transition).ToList ().ForEach (t => t.Item2.Panel.Dispose ());
```
ylatuya: There can be only one transition with this name, use First()

VAS.Core/Common/StateController.cs
```diff
@@ -243,6 +246,7 @@ public bool UnRegister (string transition)
```
ylatuya: EmptyStateStack is clearing the stack without disposing elements.

Update unit test to check that dispose is called correctly.

Weavers/Forwarder.cs
ylatuya: That is still not fixed.

ylatuya: That's still not fixed

VAS.Core/Common/StateController.cs
ylatuya: That's still not fixed

Weavers/Forwarder.cs
aameijide: I think is not possible, to delete property.PropertyType.Name.StartsWith ("ObservableCollection"). When a property is of type ObservableCollection<Something>, we have to compare with the same time (ObservableCollection<Something>) and is not possible no compare it with ObservableCollection neither ObservableCollection<T>, so I think that currently we have the best approach.

aameijide:
Ready to review again

PR #53, opened 2016-09-07
Force frame dimensions to the display width and height

ylatuya:
This fixes a bug in which some frames capturer correct
width and height in videos with PAR != 1 by fixing the width
instead of fixing the height. So the frame used in drawings
has different dimensions when taken from the player or from
the frames capturer.


dfernandezramos:
👍

PR #56, opened 2016-09-14
RA-92/RA-91: MVVMC select Projects and Delete Projects

vmartos:


VAS.Tests/Core/Common/TestStateController.cs
```diff
@@ -158,20 +176,24 @@ public void TestUnRegister_WithOverwrittenTransitions ()
...
+			Assert.AreEqual ("Transition1", lastTransition);
 			moveTransition = await sc.MoveTo ("Transition2", null);
 			Assert.IsTrue (moveTransition);
+			Assert.AreEqual ("Transition2", lastTransition);
 			moveTransition = await sc.MoveToModal ("newModalTransition", null);
 			Assert.IsTrue (moveTransition);
+			Assert.AreEqual ("newModalTransition", lastTransition);
 			moveTransition = await sc.MoveBackTo ("Transition1");
 			Assert.IsTrue (moveTransition);
+			Assert.AreEqual ("Transition1", lastTransition);
```
vguzmanp: The `MoveTo`, `MoveBackTo` and `MoveModal` asserts should be in separate tests

VAS.Core/Common/StateController.cs
```diff
@@ -57,33 +57,38 @@ public StateController ()
...
-				Log.Debug ("Moving failed because panel " + state.Panel.PanelName + " cannot move.");
+				IScreenState state = destination [transition] ();
+				bool ok = await state.PreTransition (data);
+				if (ok) {
+					await PushNavigationState (transition, state);
+				} else {
+					Log.Debug ("Moving failed because panel " + state.Panel.PanelName + " cannot move.");
+				}
+				return ok;
+			} catch (Exception ex) {
```
ylatuya: You don't really know what can fail in a transition. This try/catch is merely for debugging purpose since exceptions were simply ignored and not even printed. If we need it, we can even display a warning to the user, instead of just bailing out

VAS.Tests/Core/Common/TestStateController.cs
ylatuya: The test was already desgined like this, I am just adding extra checks to the ones already there. Plus we have already discussed that some tests can't strictly follow the AAA pattern when you need to assert intermediate states.

vguzmanp: As I understand unit testing, you should test only one thing per test. The AAA pattern is just a way to achieve that.

This test is testing several things, so to try and follow the pattern, for each of the things tested it should have an "Arrange" section that prepares the test and goes through the intermediate states, just to leave it ready to execute what you really want to test (and Assert over it).

I think I wasn't here when you discussed this, is the discussion recorded / noted anywhere?

VAS.Core/Common/StateController.cs
```diff
@@ -271,14 +271,18 @@ Task<bool> PushNavigationState (string transition, IScreenState state)
...
 		{
-			navigationStateStack [navigationStateStack.Count - 1].ScreenState.Dispose ();
+			bool ret = true;
+			IScreenState screenToPop = navigationStateStack [navigationStateStack.Count - 1].ScreenState;
 			navigationStateStack.RemoveAt (navigationStateStack.Count - 1);
 			NavigationState lastState = LastNavigationState ();
 			if (lastState != null) {
 				App.Current.EventsBroker.Publish (new NavigationEvent { Name = lastState.Name });
-				return App.Current.Navigation.LoadNavigationPanel (lastState.ScreenState.Panel);
+				ret = App.Current.Navigation.LoadNavigationPanel (lastState.ScreenState.Panel).Result;
```
vguzmanp: `.Result` is synchronous, shouldn't we `await` this?

vmartos: .Result blocks execution until it has the result, so the behaviour is similar than using await

vguzmanp: No, `Result` is the same as using `.Wait()`. `await` doesn't block the current thread, it "pauses" the current execution status of the method and continues running elsewhere. Then, when the task is finished, it resumes the method where it was.

PR #57, opened 2016-09-15
Update couchbase recipe to 1.3.1 - VAS-24

vguzmanp:


VAS.DB/packages.config
```diff
@@ -1,10 +1,12 @@
...
 <packages>
-  <package id="Couchbase.Lite.LM" version="1.2.1.1" targetFramework="net45" />
-  <package id="Couchbase.Lite.Storage.SystemSQLite.LM" version="1.2.1.1" targetFramework="net45" />
+  <package id="Couchbase.Lite.LM" version="1.3.1.1" targetFramework="net45" />
+  <package id="Couchbase.Lite.Storage.SystemSQLite.LM" version="1.3.1.1" targetFramework="net45" />
   <package id="Newtonsoft.Json" version="7.0.1" targetFramework="net45" />
   <package id="SharpZipLib.Portable" version="0.86.0.0003" targetFramework="net45" />
-  <package id="SQLitePCL.raw" version="0.8.6" targetFramework="net45" />
-  <package id="SQLitePCL.ugly" version="0.8.6" targetFramework="net45" />
+  <package id="SQLitePCL.native.sqlite3.v110_xp" version="0.9.1" targetFramework="net45" />
```
ylatuya: This provides the native built of sqlite for windows platforms, which is not cross-platform. We already ship sqlite within cerbero so it shouldn't be needed

vguzmanp: NuGet added that one as a dependency of a package, I think it was `SQLitePCL.raw`. Let me check it.

ylatuya: There are more comments about this here https://github.com/ericsink/SQLitePCL.raw/issues/66
Apparently if we want to use a custom build it needs to be named esqlite3.dll, we need to check how this new package behaves in all platforms so that the sqlite dll is imported correctly.

VAS.DB/VAS.DB.Net45.csproj
```diff
@@ -43,25 +43,30 @@
       <HintPath>$(SolutionDir)\packages\Stateless.2.5.61.0\lib\portable-net40+sl50+win+wp80+MonoAndroid10+xamarinios10+MonoTouch10\Stateless.dll</HintPath>
     </Reference>
     <Reference Include="SQLitePCL.raw">
-      <HintPath>$(SolutionDir)\packages\SQLitePCL.raw.0.8.6\lib\net45\SQLitePCL.raw.dll</HintPath>
-      <Package>couchbase-lite</Package>
+      <HintPath>..\..\packages\SQLitePCL.raw.0.9.1\lib\net45\SQLitePCL.raw.dll</HintPath>
```
ylatuya: Use $(SolutionDir) everywhere

VAS.DB/packages.config
ylatuya: We can leave it there, I don't think it will hurt since we package things ourselves, but it's a bit awkward adding a dependency to a nugget package that's not cross-platform

VAS.DB/VAS.DB.Net45.csproj
vguzmanp: Fixed, literally everywhere :smile

VAS.DB/packages.config
vguzmanp: I will test it on Windows, to see how it behaves

PR #60, opened 2016-09-21
RA-124

aoller:


VAS.UI.Gtk2/UI/Common/TreeViewBase.cs
```diff
@@ -90,7 +90,9 @@ void ViewModelCollectionChanged (object sender, NotifyCollectionChangedEventArgs
 				break;
 
 			case NotifyCollectionChangedAction.Reset:
-				ClearSubViewModelListeners (viewModel.ViewModels);
+				foreach (VM item in dictionaryStore.Keys) {
+					RemoveSubViewModelListener (item);
+				}
```
vguzmanp: Shouldn't `dictionaryStore`and `viewModel.ViewModels` store the same info?

vguzmanp: This is the same as `ClearSubViewModelListeners (dictionaryStore.Keys)`, isn't it?

PR #61, opened 2016-09-22
Fix database directory in CouchBaseStorage.Backup method

aoller:


VAS.DB/CouchbaseStorage.cs
```diff
@@ -188,14 +190,17 @@ public void Fill (IStorable storable)
...
 						StorableNode node;
 						ObjectChangedParser parser = new ObjectChangedParser ();
 						parser.Parse (out node, t, Serializer.JsonSettings);
 
 						if (!forceUpdate) {
 							Update (node);
-						} else {
+						} 
+						if (forceUpdate || documentUpdated) {
+							Info.LastModified = DateTime.UtcNow;
```
ylatuya: This change is wrong. The if/else block shouldn't be changed and just store the StorageInfo after that if there was any change in the DB

VAS.DB/CouchbaseStorage.cs
```diff
@@ -112,11 +113,12 @@ public bool Backup ()
 				using (FileStream fs = new FileStream (outputFilename, FileMode.Create, FileAccess.Write, FileShare.None)) {
 					using (Stream gzipStream = new GZipOutputStream (fs)) {
 						using (TarArchive tarArchive = TarArchive.CreateOutputTarArchive (gzipStream)) {
-							AddDirectoryFilesToTar (tarArchive, Path.Combine (App.Current.DBDir, storageName + ".cblite2"), true);
+							AddDirectoryFilesToTar (tarArchive, Path.Combine (db.Manager.Directory, storageName + ".cblite2"), true);
 						}
 					}
 				}
 				LastBackup = DateTime.UtcNow;
+				Store (Info);
```
ylatuya: This will deadlock

PR #65, opened 2016-09-27
Lon 779

ylatuya:


VAS.Tests/Services/TestCoreEventsManager.cs
```diff
@@ -0,0 +1,73 @@
...
+
+			var mtkMock = new Mock<IMultimediaToolkit> ();
+			mtkMock.Setup (m => m.GetFramesCapturer ()).Returns (capturerMock.Object);
+			App.Current.MultimediaToolkit = mtkMock.Object;
+
+			controller = new CoreEventsManager ();
+			project = Utils.CreateProject (true);
+			controller.Start ();
+			App.Current.EventsBroker.Publish (new OpenedProjectEvent {
+				Project = project, ProjectType = ProjectType.FileProject, Filter = new EventsFilterDummy (project),
```
aameijide: Is easier to use a Mock rather than create a dummy class, the code will look like this:
			Mock<EventsFilter> eventsFilterMock = new Mock<EventsFilter> (project);
			eventsFilterMock.Setup (x => x.Update ());
			App.Current.EventsBroker.Publish (new OpenedProjectEvent {
				Project = project, ProjectType = ProjectType.FileProject, Filter = eventsFilterMock.Object,
			});

Notice that Update method is configured instead of the classes inside Update method. As it is a unit test, we don't want to test eventsFilters code.

VAS.Tests/Utils.cs
```diff
@@ -79,6 +79,27 @@ public void Raise (string name)
 		}
 	}
 
+	public class EventsFilterDummy : EventsFilter
+	{
```
aameijide: This class is unnecessary. First of all, we can mock EventsFilter in our test and secondly we have another EventsFilterDummy class in line 124.

VAS.Tests/Utils.cs
```diff
@@ -245,6 +275,7 @@ public static Project CreateProject (bool withEvents = true)
 		{
 			TimelineEvent pl;
 			Project p = new ProjectDummy ();
+			p.Dashboard = DashboardDummy.Default ();
```
aameijide: Maybe we can also mock Dashboard here, but I'm afraid that if CreateProject is called from different projects we cannot mock different for each one.

VAS.UI.Gtk2/UI/Menu/MenuHelpers.cs
```diff
@@ -0,0 +1,112 @@
...
+
+		/// <summary>
+		/// Fills the menu item "Add to playlist" with all the playlist options
+		/// </summary>
+		/// <param name="addToPlaylistMenu">Add to playlist menu.</param>
+		/// <param name="project">Project.</param>
+		/// <param name="events">Timeline events.</param>
+		static public void FillAddToPlaylistMenu (MenuItem addToPlaylistMenu, Project project, IEnumerable<TimelineEvent> events)
+		{
+			if (events.Count () == 0) {
```
ylatuya: Fixed

VAS.Tests/Services/TestCoreEventsManager.cs
ylatuya: That's not the correct way of using Mock, Mock should only be used to mock interfaces since it can only mock virtual functions.

VAS.Tests/Utils.cs
ylatuya: removed

VAS.Tests/Utils.cs
ylatuya: Mock shouldn't be used here either.

VAS.Tests/Services/TestCoreEventsManager.cs
aameijide: Ok, I didn't know that Moq only works with virtual methods :-( Maybe we should considerate use always a interface for each testable class as a future work (not for now).

PR #73, opened 2016-10-04
Add App.Current.StateController.EmptyStateStack () on close app

aoller:


VAS.UI.Gtk2/GUIToolkitBase.cs
```diff
@@ -114,6 +114,7 @@ public virtual IPlayerView GetPlayerView ()
 
 		public virtual void Quit ()
 		{
+			App.Current.StateController.EmptyStateStack ().GetAwaiter ().GetResult ();
```
ylatuya: The application should not quit if EmptyStateStack return true. Make this IGUITookit.Quit API async if needed, since for other platforms we will need it too.

ylatuya: The application should not quit if EmptyStateStack return true. Make this IGUITookit.Quit API async if needed, since for other platforms we will need it too.

VAS.UI.Gtk2/GUIToolkitBase.cs
```diff
@@ -112,9 +112,9 @@ public virtual IPlayerView GetPlayerView ()
 
 		public abstract bool SelectMediaFiles (MediaFileSet fileSet);
 
-		public virtual void Quit ()
+		public virtual async Task Quit ()
 		{
-			App.Current.StateController.EmptyStateStack ().GetAwaiter ().GetResult ();
+			await App.Current.StateController.EmptyStateStack ();
```
ylatuya: The check for the return value is still missing, if some state is cancelling a transition we need to handle it here

PR #80, opened 2016-10-11
Vas 15

ylatuya:


VAS.Core/Common/Utils.cs
```diff
@@ -41,11 +41,13 @@ public static string SanitizePath (string path, params char[] replaceChars)
 		static public string SysInfo {
 			get {
 				return string.Format (
-					"Running LongoMatch {0} build:\"{1}\" OS:\"{2}\" OS Version:\"{3}\"",
+					"Running LongoMatch {0} build:\"{1}\" OS:\"{2}\" OS Version:\"{3}\" Device ID:\"{4}\"",
```
ylatuya: good point

VAS.Core/DeviceUtils.cs
```diff
@@ -0,0 +1,188 @@
...
+				line = proc.StandardOutput.ReadLine ();
+				break;
+			}
+
+			try {
+				return Guid.Parse (line);
+			} catch (Exception ex) {
+				Log.Exception (ex);
+				return Guid.Empty;
+			}
```
ylatuya: I wrote it, is it that bad? :P

vguzmanp: :smile: No, it's just something so useful and so low-level that I thought someone would have written it before.

Also I don't understand half of it 😄

PR #89, opened 2016-10-18
Use the new installation prefix for mono 4.6

turran:
We need to also look for libraries on a path like lib/mono/4.0-api/


vmartos:
👍

PR #90, opened 2016-10-18
Rename every reference to libcesarplayer

turran:
It has been removed from the Makefile.am, the .csproj
and every .cs file that references to it, along with
the correct .dll.config configurations

### DIFF
--- a/libvas/Makefile.am
+++ b/libvas/Makefile.am
@@ -4,7 +4,7 @@
 AM_CPPFLAGS = \
 	-DPACKAGE_SRC_DIR=\""$(srcdir)"\" \
 	-DPACKAGE_DATA_DIR=\""$(datadir)"\" \
-	$(CESARPLAYER_CFLAGS)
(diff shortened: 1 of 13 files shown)


libvas/Makefile.am
```diff
@@ -52,11 +52,11 @@ libcesarplayer_la_SOURCES = \
 	lgm-utils.m\
 	lgm-utils.h
 
-libcesarplayer_la_LDFLAGS = \
+libvas_la_LDFLAGS = \
 	$(CESARPLAYER_LIBS)
```
vmartos: Maybe we can rename CESARPLAYER_LIBS to VASPLAYER_LIBS?
Also CESARPLAYER_CFLAGS can be renamed to VASPLAYER_CFLAGS.

turran: Agreed, should be fixed on the new commit

vmartos:
👍

PR #94, opened 2016-10-19
Find the packages on the correct path

turran:
The path is case sensitive on linux and thus we need to
specify the correct path


vmartos:
👍

PR #95, opened 2016-10-19
Ra 110

vguzmanp:


VAS.Core/Common/StateController.cs
```diff
@@ -131,7 +131,7 @@ public StateController ()
 				return ok;
 			} catch (Exception ex) {
 				Log.Exception (ex);
-				return false;
+				throw;
```
vguzmanp: The problem is that this is swallowing the exceptions from the stateController itself.For example, if the first transition errors because it's not registered, you end up with an empty screen, no error message, and nothing to do.
I think it should throw, if the state wants to return false it can catch his own exception.

VAS.Core/Filters/Predicate.cs
```diff
@@ -0,0 +1,269 @@
...
+
+	/// <summary>
+	/// This is a composite predicate.
+	/// This predicate applies the OR operation to all the predicates it contains.
+	/// </summary>
+	public class OrPredicate<T> : CompositePredicate<T>
+	{
+		Expression<Func<T, bool>> expression = PredicateBuilder.False<T> ();
+
+		public override Expression<Func<T, bool>> Expression {
```
vguzmanp: Fixed

VAS.Core/Filters/Predicate.cs
```diff
@@ -0,0 +1,269 @@
...
+		public Expression<Func<T, bool>> Expression {
+			get;
+			set;
+		} = (a) => true;
+
+		public bool Active { get; set; } = true;
+
+		public bool Filter (T ev)
+		{
+			return Active && Expression.Compile ().Invoke (ev);
```
vguzmanp: Fixed

VAS.Core/Common/StateController.cs
vguzmanp: Removed it from the PR for the moment

PR #96, opened 2016-10-20
RA-197

aoller:


VAS.Core/App.cs
```diff
@@ -57,6 +57,8 @@ public abstract class App
 			}
 		}
 
+		public string SOFTWARE_NAME { get; private set; }
```
ylatuya: VAS.App has already a SoftwareName property.

VAS.UI.Gtk2/UI/Component/MediaFileChooser.cs
```diff
@@ -0,0 +1,202 @@
+﻿//
+//  Copyright (C) 2016 Fluendo S.A.
```
ylatuya: Keep the original copyright

VAS.UI.Gtk2/UI/Component/MediaFileChooser.cs
```diff
@@ -0,0 +1,202 @@
...
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using System;
+using VAS.Core;
+using VAS.Core.Common;
+using VAS.Core.Store;
+
+namespace VAS.UI.UI.Component
```
ylatuya: Wrong namespace

VAS.UI.Gtk2/UI/Dialog/VideoEditionProperties.cs
```diff
@@ -0,0 +1,142 @@
+﻿//
+//  Copyright (C) 2016 Fluendo S.A.
```
ylatuya: Keep the original copyright

VAS.UI.Gtk2/UI/Menu/MenuHelpers.cs
```diff
@@ -107,5 +108,34 @@ static public void FillExportToVideoFileMenu (MenuItem exportMenu, Project proje
...
 			exportMenu.SetLabel (label);
 		}
+
+		/// <summary>
+		/// Fills the menu item "Render" with all selected PlayLists and Events
+		/// </summary>
+		/// <param name="addToRenderMenu">Add to render menuItem.</param>
+		/// <param name="events">Timeline events to render.</param>
+		//FIXME: Convert this to ViewModels (both Playlist & TimelineEvent)
+		static public void FillAddToRenderMenu (MenuItem addToRenderMenu, IEnumerable<TimelineEvent> events)
```
ylatuya: Isn't this the same as FillExportToVideoFileMenu  ?

ylatuya: Why you can't use it? The menus is exactly the same.

ylatuya: The only thing that changes is the name, one is Render the other one is "Export to video file" and you can use a parameter for that

ylatuya: Also you are calling addToRenderMenu.Activated += each time you fill the menu.
So you should use MenuHelps.EmitRenderPlaylist to set the activated callback the render menu and FillExportToVideoFileMenu

VAS.UI.Gtk2/UI/Menu/PlaysMenu.cs
```diff
@@ -108,8 +114,8 @@ void CreateMenu ()
 				);
 			};
 			Add (del);
-
-			ShowAll ();
+			if (!calledFromChild)
```
ylatuya: Why do you need a to do this conditionally ?

VAS.UI.Gtk2/UI/Common/PlaylistTreeView.cs
```diff
@@ -93,13 +101,8 @@ protected override void HandleTreeviewRowActivated (object o, RowActivatedArgs a
...
 		protected override void ShowMenu ()
 		{
-//			menu = new Menu ();
-//			MenuItem item = new MenuItem ();
-//			menu.Add (item);
-//
-//			// TODO: Fill with the menu items needed
-//
-//			menu.Popup ();
+			var elements = ViewModel.Selection [0].Model.Elements.OfType<PlaylistPlayElement> ().Select (e => e.Play);
```
ylatuya: You are removing elements from the playlist, using just PlaylistPlayElement. Playlists should have their own menu, which is different from the events menu, since you are dealing with different type of objects.

VAS.UI.Gtk2/UI/Menu/PlaysMenu.cs
aoller: In order that if it is called from a derived class, the responsibility of calling ShowAll() would be in the caller.

VAS.UI.Gtk2/UI/Menu/MenuHelpers.cs
```diff
@@ -101,10 +102,17 @@ static public void FillAddToPlaylistMenu (MenuItem addToPlaylistMenu, IEnumerabl
 		/// <param name="exportMenu">Export menu.</param>
 		/// <param name="project">Project.</param>
 		/// <param name="events">Timeline events.</param>
-		static public void FillExportToVideoFileMenu (MenuItem exportMenu, Project project, IEnumerable<TimelineEvent> events)
+		static public void FillExportToVideoFileMenu (MenuItem exportMenu, Project project, IEnumerable<TimelineEvent> events,
```
ylatuya: I think this could look nicer if instead of isRender you just pass the string with the text. Than SportsPlaysMenu after calling FillExportToVideoFileMenu would call this to perform the extra check, which is only difference.
exportMenu.Visible |= project.ProjectType != ProjectType.FakeCaptureProject;

PR #98, opened 2016-10-20
Ra 127 bis

aameijide:
Move some methods to templates and add validations

### DIFF
--- a/VAS.Services/Controller/TemplatesController.cs
+++ b/VAS.Services/Controller/TemplatesController.cs
@@ -326,7 +326,6 @@ namespace VAS.Services.Controller
 		{
 			TModel template = evt.Object;
 			bool force = evt.Force;
-
 			TViewModel templateViewmodel = ViewModel.ViewModels.FirstOrDefault (vm => vm.Model.Equals (template));
 			bool savedOk = false;
 
--- a/VAS.Services/Controller/TemplatesController.cs
+++ b/VAS.Services/Controller/TemplatesController.cs
@@ -20,7 +20,9 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Collections.Specialized;
 using System.ComponentModel;
+using System.Dynamic;
 using System.Linq;
+using System.Threading.Tasks;
 using VAS.Core;
 using VAS.Core.Common;
 using VAS.Core.Events;
@@ -114,6 +116,12 @@ namespace VAS.Services.Controller
 
 		protected string Extension { get; set; }
 
+		/// <summary>
+		/// Gets or sets the name of the transition used in HandleOpen
+		/// </summary>
+		/// <value>The name of the open transition.</value>
+		protected string OpenTransitionName { get; set; }
+
 		protected ITemplateProvider<TModel> Provider {
 			get {
 				return provider;
@@ -122,6 +130,8 @@ namespace VAS.Services.Controller
 				provider = value;
 			}
 		}
+
+		protected abstract bool ShouldCloseOnSave { get; }
 
 		#region IController implementation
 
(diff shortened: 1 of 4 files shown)


VAS.Services/Controller/TemplatesController.cs
```diff
@@ -139,6 +143,7 @@ public virtual void Start ()
 			App.Current.EventsBroker.Subscribe<ExportEvent<TModel>> (HandleExport);
 			App.Current.EventsBroker.Subscribe<ImportEvent<TModel>> (HandleImport);
 			App.Current.EventsBroker.Subscribe<UpdateEvent<TModel>> (HandleSave);
+			App.Current.EventsBroker.Subscribe<OpenEvent<TModel>> (async (evt) => await HandleOpen (evt));
```
ylatuya: Why don't you use the same pattern has for the other events ?

VAS.Services/Controller/TemplatesController.cs
```diff
@@ -155,6 +160,7 @@ public virtual void Stop ()
 			App.Current.EventsBroker.Unsubscribe<ExportEvent<TModel>> (HandleExport);
 			App.Current.EventsBroker.Unsubscribe<ImportEvent<TModel>> (HandleImport);
 			App.Current.EventsBroker.Unsubscribe<UpdateEvent<TModel>> (HandleSave);
+			App.Current.EventsBroker.Unsubscribe<OpenEvent<TModel>> (async (evt) => await HandleOpen (evt));
```
ylatuya: This is not going to work since you are trying to unsubscribe from a lambda expresion

VAS.Services/Controller/TemplatesController.cs
```diff
@@ -303,12 +316,16 @@ public IEnumerable<KeyAction> GetDefaultKeyActions ()
 			ObservableCollection<TModel> templates = evt.Object;
 
 			if (templates != null) {
-				string msg = ConfirmDeleteListText;
-				if (await App.Current.Dialogs.QuestionMessage (msg, null)) {
-					foreach (TModel template in templates) {
-						Provider.Delete (template);
+				if (templates.Count () == 1) {
```
ylatuya: Why do you need a specific behavior when it's 1?

VAS.Services/Controller/TemplatesController.cs
```diff
@@ -303,12 +316,16 @@ public IEnumerable<KeyAction> GetDefaultKeyActions ()
...
+					HandleDelete (new DeleteEvent<TModel> () { Object = evt.Object.FirstOrDefault () });
+				} else {
+					string msg = ConfirmDeleteListText;
+					if (await App.Current.Dialogs.QuestionMessage (msg, null)) {
+						foreach (TModel template in templates) {
+							Provider.Delete (template);
+						}
+						viewModel.Select (viewModel.Model.FirstOrDefault ());
```
ylatuya: You are now selecting only when you delete many

VAS.Services/ViewModel/TemplateManagerViewModel.cs
```diff
@@ -96,13 +97,24 @@ public void New ()
...
+		/// <summary>
 		/// Command to delete the currently loaded template.
 		/// </summary>
-		public virtual void Delete ()
+		public void Delete ()
 		{
-			TViewModel templateVM = Selection.FirstOrDefault ();
-			if (templateVM != null && templateVM.Model != null) {
-				App.Current.EventsBroker.Publish (new DeleteEvent<TModel> { Object = templateVM.Model });
+			if (Selection != null && !Selection.Any (x => x.Model == null)) {
```
ylatuya: That's not correct, the previous implementation was only sending the event if templateVM.Model != null, no you include them all.
You should be doing:
ObservableCollection<TModel> objects = new ObservableCollection<TModel> (  Selection.Where (x => x.Model != null).Select (x => x.Model));
if (objects.Any()) {
send event
}

VAS.Services/Controller/TemplatesController.cs
aameijide: Ok, I didn't see them, I change the others also

VAS.Services/Controller/TemplatesController.cs
aameijide: Select was also made in HandleDelete. Anyway, these methods are going to be merge, so it will be to both.

PR #99, opened 2016-10-24
RA-198

aoller:


VAS.Services/ViewModel/RenderingStateBarVM.cs
```diff
@@ -0,0 +1,55 @@
...
+//
+using System;
+using VAS.Core.Interfaces.GUI;
+using VAS.Core.Interfaces.MVVMC;
+using VAS.Core.MVVMC;
+
+namespace VAS.Services.ViewModel
+{
+	public class RenderingStateBarVM : BindableBase, IViewModel, IRenderingStateBar
+	{
```
aameijide: Document public methods/properties

VAS.UI.Gtk2/UI/Component/RenderingStateBarView.cs
```diff
@@ -0,0 +1,85 @@
...
+
+		void SyncVMValues ()
+		{
+			if (viewModel == null)
+				return;
+
+			Visible = viewModel.JobRunning;
+			statebutton.Label = viewModel.Text;
+			progressbar.Text = viewModel.ProgressText;
+			progressbar.Fraction = viewModel.Fraction;
```
aameijide: if you set this values in the VM this method is not necessary. I mean, you could do this at VM:
public bool JobRunning { get { return Model.Visible; } set { Model.Visible = value; } } and the same with the others values

aameijide: Has we talked, I was bas understanding this, so yours is correct.

PR #100, opened 2016-10-24
LON-836: Add a EmptyTag static variable and show Event if Empty

vmartos:


VAS.Core/Filters/EventsFilter.cs
```diff
@@ -222,7 +222,8 @@ protected virtual bool IsVisibleByCategory (TimelineEvent play)
 				return true;
 
 			List<Tag> tags = eventsFilter [play.EventType];
-			if (tags.Count == 0 || tags.Intersect (play.Tags).Any ()) {
+			if (tags.Count == 0 || tags.Intersect (play.Tags).Any () ||
+				(!play.Tags.Any () && tags.Contains (Tag.EmptyTag))) {
```
ylatuya: Can you add a test for this?

VAS.UI.Multimedia.Gtk2/UI/PlayerCapturerBin.cs
```diff
@@ -75,7 +75,9 @@ public void SetViewModel (object viewModel)
 
 		protected override bool OnExposeEvent (Gdk.EventExpose evnt)
 		{
-			ShowPlayer ();
+			if (mode == PlayerViewOperationMode.LiveAnalysisReview) {
```
ylatuya: Shouldn't we forward expose events also to the player if it's visible ?

PR #101, opened 2016-10-26
Now PreTransitions calls new Method Initialize

vmartos:
This is done to prevent casting problems with overriden PreTransition
when calling the base method, now if some state overrides PreTransition
It should call Initiaze instead of base.PreTransition()

### DIFF
--- a/VAS.Services/State/ScreenState.cs
+++ b/VAS.Services/State/ScreenState.cs
@@ -109,6 +109,11 @@ namespace VAS.Services.State
 
 		public virtual Task<bool> PreTransition (dynamic data)
 		{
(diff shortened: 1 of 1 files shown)


VAS.Services/State/ScreenState.cs
```diff
@@ -107,7 +107,7 @@ public virtual Task<bool> PostTransition ()
 			return AsyncHelpers.Return (true);
 		}
 
-		public virtual Task<bool> PreTransition (dynamic data)
+		protected Task<bool> Initialize (dynamic data)
```
ylatuya: wrong place for protected method

PR #105, opened 2016-10-28
Fix linq Max function raises exception on empty collections

aoller:


vmartos:
Can we add a unit test in TestPlayerController to verify this don't happen again?

VAS.Tests/Services/TestPlayerController.cs
```diff
@@ -1570,6 +1576,19 @@ public void TestPlayerControllerRefreshMediaFileSetWhenItChanges ()
 
 		}
 
+		[Test ()]
+		public void TestLoadEventWithoutCamerasConfig ()
+		{
+			try {
+				player = new PlayerController (true);
+				(player as IController).SetViewModel (playerVM);
```
vmartos: player = new PlayerController (true);
(player as IController).SetViewModel (playerVM);
this two lines are not needed

vmartos:
👍

PR #108, opened 2016-10-31
RA-244

aoller:


VAS.Services/State/ScreenState.cs
```diff
@@ -108,8 +108,6 @@ public virtual Task<bool> PostTransition ()
 			foreach (IController controller in Controllers) {
 				controller.Stop ();
 			}
-			RootControllers.Clear ();
-			Controllers.Clear ();
```
vguzmanp: This should be modified so that `Controllers` ends up with the same as `RootControllers`

This way, `PreTransition` will recreate the non-root controllers, and start all of them.

PR #115, opened 2016-11-07
Ra 176: Fix Seeks in PlayerController KeyboardInput, In the TimeRule.

vmartos:


VAS.Services/PlayerController.cs
```diff
@@ -865,22 +865,46 @@ IEnumerable<KeyAction> IController.GetDefaultKeyActions ()
 				new KeyAction (
 					"PLAYER_SEEK_LEFT_SHORT",
 					App.Current.Keyboard.ParseName ("Left"),
-					() => Seek (CurrentTime + new Time { TotalSeconds = -SHORT_SEEK_SECONDS })
+					() => {
+						Time seekTime = CurrentTime + new Time { TotalSeconds = -SHORT_SEEK_SECONDS };
```
ylatuya: PlayerController has the PerformStep function for that

VAS.Drawing/CanvasObjects/Timeline/CameraObject.cs
```diff
@@ -372,20 +373,36 @@ protected virtual void HandleChangeVideoSizeEvent (ChangeVideoSizeEvent e)
...
+				if (e.SeekType == SeekType.StepUp) {
+					e.player.SeekToNextFrame ();
+					TimeNode.Stop = e.player.CurrentTime;
+				} else if (e.SeekType == SeekType.StepDown) {
+					e.player.SeekToPreviousFrame ();
+					TimeNode.Stop = e.player.CurrentTime;
+				} else {
+					Time stopTime = TimeNode.Stop + e.Time;
+					if (stopTime <= MaxTime) {
+						// Almost 1s must be showed in the timeline
```
ylatuya: The code is very simliar to the one you have when moving the Left side, you can put it in the same fuction with the upper and lower bounds checks.

VAS.Services/PlayerController.cs
vmartos: The difference is that the PerformStep only checks for <0 and duration, but we need to check for the loadedSegment start and stop, because if we are in a event we don't wan't to seek to a time > event.Stop

VAS.Tests/Services/TestPlayerController.cs
```diff
@@ -1587,6 +1602,56 @@ public void TestLoadEventWithoutCamerasConfig ()
...
+			currentTime = new Time (5000);
+
+			player.LoadCameraEvent (cameraEvent, currentTime - cameraEvent.Start, false);
+			player.LoadEvent (evt2, new Time (0), false);
+			currentTime = new Time (7000);
+			player.UnloadCurrentEvent ();
+			player.Seek (new Time (1000), true);
+
+			playerMock.Verify (p => p.Seek (currentTime, true, false), Times.Once ());
+			playerMock.Verify (p => p.Seek (new Time (3000), true, false), Times.Once ());
```
ylatuya: Can you add some comments for this test? It took me a while to understand you where testing the second seek to 1000 :)

PR #116, opened 2016-11-07
RA-128

dfernandezramos:


VAS.Services/Controller/PlaylistController.cs
```diff
@@ -32,7 +32,12 @@ public class PlaylistController : DisposableBase, IController
 	{
 		PlaylistCollectionVM viewModel;
 
-		public PlaylistController (IPlayerViewModel playerVM)
+		public IVideoPlayerViewModel PlayerVM {
+			get;
+			set;
+		}
+
```
vguzmanp: Constructor should go before properties

VAS.Core/Common/Cloner.cs
```diff
@@ -29,15 +30,19 @@ public static T Clone<T> (this T source, SerializationType type = SerializationT
...
 
 			Stream s = new MemoryStream ();
 
 			// Binary deserialization fails in mobile platforms because of
 			// https://bugzilla.xamarin.com/show_bug.cgi?id=37300
-			#if OSTYPE_ANDROID || OSTYPE_IOS
+#if OSTYPE_ANDROID || OSTYPE_IOS
 			type = SerializationType.Json;
-			#endif
+#endif
```
vguzmanp: I hate when xamarin does this...

VAS.Core/Common/Cloner.cs
```diff
@@ -29,15 +30,19 @@ public static T Clone<T> (this T source, SerializationType type = SerializationT
...
+
+			if (source is BindableBase) {
+				type = SerializationType.Json;
```
vguzmanp: Most of our objects are BindableBase (or will be). What does this imply?

VAS.Core/Store/Templates/Team.cs
```diff
@@ -98,18 +97,8 @@ protected virtual void Dispose (bool disposing)
...
-				if (list != null) {
-					list.CollectionChanged -= PlayersChanged;
-				}
-				list = value;
-				if (list != null) {
-					list.CollectionChanged += PlayersChanged;
-				}
-			}
+			get;
+			set;
```
vguzmanp: Now PlayersChanged is never called. Shouldn't we call it on `List.CollectionChanged`?

VAS.Core/Common/Cloner.cs
aameijide: This solve problems with cloned items that comes from DB. It indicates that the serialization type must be Json, like the one we have in DB.

We we're using that has a parameter in HandleSelectionChange from ProjectsController or TemplatesController, so this made that we don't have to put this parameters each time we have a Bindable base

vguzmanp: So why not use this always?

VAS.Core/Store/Templates/Team.cs
aameijide: It's no longer needed, so I remove it

PR #118, opened 2016-11-07
RA-250: Save playlist when it change

aameijide:


aameijide:
DB View was already created.

VAS.Tests/Services/TestPlaylistController.cs
```diff
@@ -38,13 +39,21 @@ public class TestPlaylistController
...
 		public void FixtureSetup ()
 		{
 			mockPlayerController = new Mock<IPlayerViewModel> ();
 			mockGuiToolkit = new Mock<IGUIToolkit> ();
+
+			storageManagerMock = new Mock<IStorageManager> ();
+			storageManagerMock.SetupAllProperties ();
+			storageMock = new Mock<IStorage> ();
+			storageManagerMock.Object.ActiveDB = storageMock.Object;
+			App.Current.DatabaseManager = storageManagerMock.Object;
```
ylatuya: Why is this change needed at all ?

aameijide: Playlist are saved when are created or modified, so we need to mock Database in test. If not, TestAddSamePlaylist and TestNewPlaylist fails

PR #119, opened 2016-11-08
RA-253

aoller:


VAS.Tests/Services/TestScreenState.cs
```diff
@@ -118,37 +116,73 @@ public void TestPreTransition ()
...
+		}
+
+		[Test ()]
+		public void TestUnloadState ()
+		{
+			// Arrange
+			TestHideState ();
+
+			// Act
+			screenState.UnloadState ();
```
dfernandezramos: Just asking: Wouldn't be nice if we implement the hidestate method into the unloadstate method in case the state hasn't been already hidden? Just to avoid unloading a state before it's hidden.

dfernandezramos:
👍

VAS.Core/Common/StateController.cs
```diff
@@ -46,6 +46,7 @@ public StateController ()
 			try {
 				Log.Debug ("Setting Home to " + transition);
 				IScreenState homeState = destination [transition] ();
+				homeState.LoadState (properties);
```
ylatuya: LoadState is not awaited and return  value is ignored

VAS.Core/Common/StateController.cs
```diff
@@ -273,6 +275,7 @@ Task<bool> PushNavigationState (string transition, IScreenState state, bool isho
 			if (!ishome) {
 				navigationStateStack.Add (new NavigationState (transition, state));
 			}
+			state.ShowState ();
```
ylatuya: ShowState is not awaited

VAS.Core/Common/StateController.cs
```diff
@@ -93,11 +94,9 @@ public StateController ()
 					state = home.ScreenState;
 				} else {
 					state = destination [transition] ();
+					state.LoadState (properties);
```
ylatuya: LoadState is not awaited and returned value  is ignored

VAS.Core/Interfaces/GUI/IScreenState.cs
```diff
@@ -54,17 +54,31 @@ public interface IScreenState : IDisposable
...
 		/// <summary>
-		/// Performs steps prior to the state transition, like initialization the View, ViewModel and Controllers.
+		/// Method called once when transition is loaded for first time.
+		/// This method creates the controllers. And receives data.
 		/// </summary>
-		/// <returns><c>True</c> if the transition was successful, <c>False</c> otherwise.</returns>
-		/// <param name="data">The parameters from a previous state.</param>
-		Task<bool> PreTransition (dynamic data);
+		/// <returns>The task result.</returns>
+		/// <param name="data">ViewModel to be used</param>
```
ylatuya: It can be a ViewModel or an ExpandoObject with other properties

VAS.Core/MVVMC/CollectionViewModel.cs
```diff
@@ -131,7 +131,10 @@ void HandleModelsCollectionChanged (object sender, NotifyCollectionChangedEventA
 				}
 				break;
 			case NotifyCollectionChangedAction.Reset:
-				ViewModels.Clear ();
+				foreach (var vm in ViewModels.ToList ()) {
+					ViewModels.Remove (vm);
```
ylatuya: You should keep ViewModels.Clear() because it send a Reset event to observers instead of one event per element deleted.

VAS.Services/Controller/PlaylistController.cs
```diff
@@ -64,7 +66,10 @@ public void Stop ()
 
 		public void SetViewModel (IViewModel viewModel)
 		{
-			this.viewModel = (PlaylistCollectionVM)viewModel;
+			if (viewModel == null) {
+				return;
```
ylatuya: If viewModel is null it should be doing something else, at least resetting the instance variable viewModel

VAS.Services/Controller/ProjectsController.cs
```diff
@@ -198,6 +206,9 @@ Task HandleNew (CreateEvent<TModel> evt)
 
 		async Task<bool> Save (TModel project, bool force)
 		{
+			if (!ShouldHandleSave) {
```
ylatuya: I don't understand this change

PR #122, opened 2016-11-09
Fixes a bug when closing new projects when playing video

vmartos:
No need to register another glib timout when setting
a new MediaFileSet

### DIFF
--- a/VAS.Drawing/CanvasObjects/Timeline/CameraObject.cs
+++ b/VAS.Drawing/CanvasObjects/Timeline/CameraObject.cs
@@ -66,11 +66,17 @@ namespace VAS.Drawing.CanvasObjects.Timeline
 			App.Current.EventsBroker.Subscribe<ChangeVideoSizeEvent> (HandleChangeVideoSizeEvent);
 		}
 
-		public void Dispose ()
+		protected override void Dispose (bool disposing)
 		{
-			stretchedAndEditingTimer.Dispose ();
-			App.Current.EventsBroker.Unsubscribe<StretchVideoEvent> (HandleStrechVideoEvent);
-			App.Current.EventsBroker.Unsubscribe<ChangeVideoSizeEvent> (HandleChangeVideoSizeEvent);
+			if (Disposed)
+				return;
+
+			base.Dispose (disposing);
+			if (disposing) {
+				stretchedAndEditingTimer.Dispose ();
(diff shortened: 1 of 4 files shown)


VAS.Drawing/CanvasObjects/Timeline/CameraObject.cs
```diff
@@ -354,7 +354,7 @@ void SetVideoTlMode ()
 		protected virtual void HandleStrechVideoEvent (StretchVideoEvent e)
 		{
 			var ev = new TimelineEvent ();
-			if (IsStretched ()) {
+			if (e.ForceEditMode || IsStretched ()) {
```
vguzmanp: I prefer to do this without a Hack, I think the `CameraObject` (or whoever contains it) should be subscribed to `LoadVideoEvent` and react by setting everything to default, etc.

The `StretchVideoEvent` should be used only to apply/unapply the current stretch

VAS.Drawing/CanvasObjects/Timeline/CameraObject.cs
```diff
@@ -373,6 +375,20 @@ protected virtual void HandleStrechVideoEvent (StretchVideoEvent e)
...
+			var ev = new TimelineEvent ();
+			ev.Start = new Time (0);
+			ev.Stop = e.mfs.Duration;
+			ev.FileSet = e.mfs;
+			ev.EventTime = ev.Duration;
+			App.Current.EventsBroker.Publish<LoadCameraEvent> (
+				new LoadCameraEvent {
+					CameraTlEvent = ev
+				});
+		}
```
vguzmanp: I still don't see this.
As I understand it, when replacing the video, the cameraObject/timeline/whoever should do the same they do the first time.

In this case, it would be
```
    public CameraObject (MediaFile mf) :
 			base (new TimeNode {
 				Start = new Time (-mf.Offset.MSeconds),
 				Stop = mf.Duration - mf.Offset, Name = mf.Name
 			})
 		{
 			VideoTLmode = VideoTimelineMode.Default;
 			mediaFile = mf;
 			DraggingMode = NodeDraggingMode.Borders;
 			SelectionMode = NodeSelectionMode.Borders;
 			ClippingMode = NodeClippingMode.LeftStrict;
 			stretchedAndEditingTimer = new SysTimer (400);
 			stretchedAndEditingTimer.Elapsed += HandleTimeOut;
```
But this code is different, why the difference?

PR #126, opened 2016-11-11
RA-266 - Show the correct menu in the timeline

vguzmanp:
If it's not a known timeline, don't show anything


aameijide:
Running test fails at: VAS.Tests.Services.TestPlayerController.TestLoadCameraEvent ()  and VAS.Tests.TestPlaylistManager.TestLoadCameraPlay_CameraTlEventIsNull ()

PR #128, opened 2016-11-14
Ra 276 range observable collection

vmartos:


VAS.Core/Common/RangeObservableCollection.cs
```diff
@@ -25,21 +24,35 @@ namespace VAS.Core.Common
...
+
+		public RangeObservableCollection (IEnumerable<T> collection) : base (collection)
+		{
+		}
+
 		public void AddRange (IEnumerable<T> items)
 		{
-			foreach (var item in items) {
-				Items.Add (item);
+			if (items.Any ()) {
```
vguzmanp: Check items for null

VAS.Core/Common/RangeObservableCollection.cs
```diff
@@ -25,21 +24,35 @@ namespace VAS.Core.Common
...
-			OnCollectionChanged (e);
 		}
 
 		public void RemoveRange (IEnumerable<T> items)
 		{
 			foreach (var item in items) {
 				Items.Remove (item);
 			}
-			NotifyCollectionChangedEventArgs e = new NotifyCollectionChangedEventArgs (NotifyCollectionChangedAction.Remove, items);
+			NotifyCollectionChangedEventArgs e = new NotifyCollectionChangedEventArgs (NotifyCollectionChangedAction.Remove, items.ToList ());
```
vguzmanp: Why the conversion to List?

VAS.Core/MVVMC/CollectionViewModel.cs
```diff
@@ -98,12 +111,10 @@ void HandleViewModelsCollectionChanged (object sender, NotifyCollectionChangedEv
 			editing = true;
 			switch (e.Action) {
 			case NotifyCollectionChangedAction.Add:
-				foreach (TViewModel viewModel in e.NewItems)
-					model.Add (viewModel.Model);
+				model.AddRange (e.NewItems.OfType<TViewModel> ().Select ((arg) => arg.Model));
```
vguzmanp: Isn't `e.NewItems` already of type `TViewModel`? The old code wouldn't work if it wasn't, would it?

VAS.Core/MVVMC/CollectionViewModel.cs
```diff
@@ -120,19 +131,17 @@ void HandleModelsCollectionChanged (object sender, NotifyCollectionChangedEventA
...
 			case NotifyCollectionChangedAction.Remove:
+				ViewModels.RemoveRange (e.OldItems.OfType<TModel> ().Select ((arg) => modelToViewModel [arg]));
 				foreach (TModel model in e.OldItems) {
-					ViewModels.Remove (modelToViewModel [model]);
 					modelToViewModel.Remove (model);
 				}
 				break;
 			case NotifyCollectionChangedAction.Reset:
+				ViewModels.Clear ();
 				foreach (var vm in ViewModels.ToList ()) {
```
vguzmanp: now you don't need the `ToList`, right?

VAS.Services/PlayerController.cs
```diff
@@ -207,6 +207,9 @@ public PlayerController (bool supportMultipleCameras = false)
 				if (StillImageLoaded) {
 					return imageLoadedTS;
 				} else {
+					if (player.CurrentTime == new Time (0)) {
+						bool breakpoint = true;
+					}
```
vguzmanp: 😃

VAS.Core/MVVMC/CollectionViewModel.cs
vguzmanp: Actually, this should be done the other way round

VAS.Core/Common/RangeObservableCollection.cs
vmartos: I don't know either :), but this solved the crash Invalid cast exception in the CollectionChanged event handlers...

vguzmanp: true, the event is declared as an `IList` (https://msdn.microsoft.com/en-us/library/system.collections.specialized.notifycollectionchangedeventargs(v=vs.110).aspx)

VAS.Core/Common/RangeObservableCollection.cs
vmartos: I don't know if we should check for null. Maybe better to let the app crash, because shouldn't be possible to do AddRange(null)

VAS.Core/MVVMC/CollectionViewModel.cs
vmartos: e.NewItems is of Type IList, so a cast should be performed with working with collections... The Old performs an implicit cast to TViewModel and that's why it worked.

VAS.Core/MVVMC/CollectionViewModel.cs
vmartos: yes!

PR #130, opened 2016-11-16
RA-168

aoller:


vguzmanp:
I remember the stretch behaviour was implemented in the PlaylistManager, or something like that, but I haven't seen it here. Is it anywhere? Is it working?

PR #136, opened 2016-11-22
Call HideState before push modal, and ShowState after pop modal

aoller:


VAS.Core/Common/StateController.cs
```diff
@@ -128,6 +128,13 @@ public StateController ()
 			try {
 				Log.Debug ("Moving to " + transition + " in modal mode");
 				IScreenState state = destination [transition] ();
+
+				bool isModal;
+				NavigationState lastState = LastState (out isModal);
```
vguzmanp: If you only need the state, I think you can call `LastState()` directly

PR #140, opened 2016-11-28
RA-318

dfernandezramos:


VAS.Core/StyleConf.cs
```diff
@@ -23,6 +23,21 @@ namespace VAS.Core.Common
 	// FIXME: Split this in LongoMatch-specific & VAS styles
 	public class StyleConf
 	{
+		public StyleConf ()
+		{
+			HomeTeamColor = Color.Red;
```
vmartos: Constructor should go after constants and before properties, Also do we need to Initialise All Properties? What happens if you delete the json containing all values?

dfernandezramos: Done.
Answering your question: Look that there is not initialising all properties, just the colors. This is in case we delete the json or something like that the app theme would be black. If we don't initialise the properties and the json is deleted the theme would be white.

vmartos:
👍

PR #143, opened 2016-12-01
Ra 260

ylatuya:


VAS.Core/Store/MediaFileSet.cs
```diff
@@ -40,6 +40,7 @@ public MediaFileSet ()
 		{
 			ID = Guid.NewGuid ();
 			IsLoaded = true;
+			VisibleRegion = new TimeNode { Start = new Time (-1), Stop = new Time (-1) };
```
vguzmanp: Why is this? I think null was cleaner

VAS.Core/Store/MediaFileSet.cs
```diff
@@ -114,6 +114,15 @@ public MediaFileSet ()
 			set;
 		}
 
+		/// <summary>
+		/// Gets or sets if the media file is stretched.
+		/// </summary>
+		/// <value>The is stretched.</value>
+		public bool IsStretched {
+			get;
```
vguzmanp: Shouldn't the getter be `VisibleRegion != null` or something like that?

VAS.Core/Store/MediaFileSet.cs
```diff
@@ -312,16 +321,16 @@ public override int GetHashCode ()
 
 		protected override void OnCollectionChanged (NotifyCollectionChangedEventArgs e)
 		{
-			if (Count != 0) {
+			if (Count != 0 && Duration != null) {
```
vguzmanp: Can Duration be null?

VAS.Core/Store/MediaFileSet.cs
ylatuya: VisibleRegion is not null when IsStretched is False.

VAS.Core/Store/MediaFileSet.cs
ylatuya: Yes, when one of the  MediaFiles is not initialized correctly.

VAS.Core/Store/MediaFileSet.cs
vguzmanp: Ok, my point still stands, shouldn't the getter be dependent on the actual value used for stretching?

ylatuya: that's the real value used

VAS.Services/VideoPlayerController.cs
```diff
@@ -288,7 +288,14 @@ public VideoPlayerController (bool supportMultipleCameras = false)
 				mode = value;
 				if (FileSet?.FirstOrDefault () != null) {
 					visibleRegion = FileSet.VisibleRegion;
+					Time currentTime = CurrentTime;
 					UnloadCurrentEvent ();
+					if (Mode == VideoPlayerOperationMode.Stretched) {
+						LoadSegment (mediafileSet, visibleRegion.Start, visibleRegion.Stop,
+									 CurrentTime.Clamp (visibleRegion.Start, visibleRegion.Stop), (float)Rate,
+									 CamerasConfig, CamerasLayout, Playing);
+						UpdateDuration ();
```
vguzmanp: In `UnloadCurrentEvent` there's also a LoadSegment, but UpdateDuration is called before it. Does the order matter?

VAS.Core/Store/MediaFileSet.cs
ylatuya: I had to do it because the MediaFileVM is creating a TimeNodeVM for VisibleRegion and the model is not updated when replacing VisibleRegion, but I can try to change it back.

VAS.Services/VideoPlayerController.cs
```diff
@@ -1421,13 +1421,35 @@ protected virtual bool Tick ()
...
 				}
 				videoTS = currentTime;
 				return true;
 			}
 		}
 
+		void UpdateDuration ()
+		{
+			if (mode == VideoPlayerOperationMode.Presentation) {
+				duration = LoadedPlaylist?.Duration;
```
vguzmanp: I think this should crash if there's no LoadedPlaylist.

VAS.Tests/Core/Store/TestMediaFileSet.cs
```diff
@@ -313,6 +314,20 @@ public void TestVisibleRegionOutOfBounds ()
...
+		{
+			MediaFileSet mfs = new MediaFileSet ();
+			mfs.Add (new MediaFile {
+				FilePath = "/videos/test.mp4",
+				Duration = new Time (20000),
+			});
+			mfs.VisibleRegion.Stop.MSeconds += 100;
+
+			Assert.AreEqual (new Time (20000), mfs.Duration);
+			Assert.IsFalse (mfs.VisibleRegion.Stop == mfs.Duration);
```
vguzmanp: Do we have a test for when it changes the duration? (`IsStretched` is true)

VAS.Tests/Services/TestVideoPlayerController.cs
```diff
@@ -1521,7 +1520,7 @@ public void TestPresentationSeekLongerThanFileset ()
 			App.Current.EventsBroker.Subscribe<LoadPlaylistElementEvent> ((e) => playlistElementSelected++);
 
 			player.Mode = VideoPlayerOperationMode.Presentation;
-			Assert.IsFalse (player.Seek (new Time (4000), true, false, false));
+			Assert.IsFalse (player.Seek (seekTime, true, false, false));
```
vguzmanp: This test was testing a seek inside the duration, and now it's testing outside it

VAS.Drawing/Widgets/Timerule.cs
```diff
@@ -305,40 +304,30 @@ public override void Draw (IContext context, Area area)
 		protected override void StartMove (Selection sel)
 		{
 			WasPlaying = PlayingState;
-			App.Current.EventsBroker.Publish<TogglePlayEvent> (
-				new TogglePlayEvent {
-					Playing = false
-				}
-			);
+			ViewModel.VideoPlayer.LoadEvent (null, false);
```
vguzmanp: Did you test this in LM? I think presentations relied on that for the seeks with the different players, etc...

VAS.Services/VideoPlayerController.cs
ylatuya: Yes, this was triggered in a unit test that set the mode before the PL, you are right and it should throw

VAS.Services/VideoPlayerController.cs
ylatuya: right, fixed

VAS.Tests/Services/TestVideoPlayerController.cs
ylatuya: It's testing a seek outside the duration: TestPresentation**SeekLongerThanFileset**
It' started like: Assert.Greater (new Time (4000), mfs.Duration);

vguzmanp: True, sorry. I said that because it was seeking to 4000, and `element0.Play.Stop` was 5000, but now it's seeking to 51000s and `element0.Play.Stop` is 6000s. I find that change strange.

VAS.Tests/Core/ViewModel/TestMediaFileSetVM.cs
```diff
@@ -0,0 +1,108 @@
...
+				Model = model,
+			};
+			model.Add (new MediaFile {
+				FilePath = "/videos/test.mp4",
+				Duration = new Time (20000),
+			});
+			model.VisibleRegion.Start = new Time (7000);
+			model.VisibleRegion.Stop = new Time (17000);
+
+			model.IsStretched = true;
```
vguzmanp: IsStretched is already true

VAS.Tests/Core/Store/TestMediaFileSet.cs
vguzmanp: I can't see it, did you push it?

PR #152, opened 2016-12-12
Check first if the key exists before adding in dictionary

vmartos:


VAS.Core/ViewModel/TimelineVM.cs
```diff
@@ -174,6 +174,9 @@ protected virtual void HandleEventTypesCollectionChanged (object sender, NotifyC
 
 		void AddTimelineEventVM (TimelineEventVM viewModel)
 		{
+			if (!eventTypeToTimeline.ContainsKey (viewModel.Model.EventType)) {
+				ViewModels.Add (new EventTypeTimelineVM { Model = viewModel.Model.EventType });
+			}
```
ylatuya: A unit test would be nice to have

PR #154, opened 2016-12-13
Define commands in the ViewModel to change the volume and the rate

ylatuya:
Leave the properties justi to reflect the current value and update the UI

### DIFF
--- a/VAS.Core/ViewModel/VideoPlayerVM.cs
+++ b/VAS.Core/ViewModel/VideoPlayerVM.cs
@@ -69,17 +69,22 @@ namespace VAS.Core.ViewModel
 			set;
 		} = true;
 
-		[PropertyChanged.DoNotNotify]
+		/// <summary>
+		/// Gets or sets the current audio volume.
+		/// This value is only used for display in the view. To change the volume use <see cref="SetVolume"/>
+		/// </summary>
+		/// <value>The new volume.</value>
 		public double Volume {
-			get {
-				return playerController.Volume;
-			}
-			set {
-				playerController.Volume = value;
-			}
+			get;
+			set;
 		}
 
-		public float Rate {
+		/// <summary>
+		/// Gets or sets the current playback rate.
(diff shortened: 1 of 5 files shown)


vguzmanp:
How is this fixing the rate? I'm not sure why we need this...

ylatuya:
In the player there are 2 different things:

1. A property in the ViewModel to reflect a current value, like Volume, Rate, CurrentTime
2. A Command to perform an action in the ViewModel, like changing the volume, the rate or seeking.

Right for the volume and the rate we are mixing both. In the case of the rate,the View is trying the change the rate using the Rate property in the ViewModel, which does nothing. In the case of the volume, setting the Volume changes the volume but there is no way to notify to the view that the volume has changed for example.
This is why it need to be split in Properties and Commands. When the rate changes (from any kind of source, it can be the View or a Key Binding), the controller updates this property and the View respond the property change updating the new value, like it's done for CurrentTime.

vguzmanp:
If we changed the VideoPlayerController to be the same as other controllers (i.e: responding to events, and modifying the VM's properties) this problem would go away, right?

ylatuya:
The problem would be the same if you use the Rate property directly, instead of a command. In ViewModel backed by a model, the model is itself emitting the property change when the value changes. In this case, we don't have any model, so the only way to notify rate chagnes is through a property in the ViewModel. That's why we need to separate the commands and the events. It's like seeking works, you have a command to perform a seek and the CurrentTime property to expose the current time to the Views

vguzmanp:
Ahhm OK.
We could add a comment explaining all that to the class.

vguzmanp:
And are there unit tests for this?

ylatuya:
It's already documented in the properties: https://github.com/fluendo/VAS/pull/154/files#diff-20918f7489682900dd453c463c7f1bf6R74

vguzmanp:
Yes, but it's something related to the whole class:

> In this case, we don't have any model

so it would be useful to document it on the class, to refer to that in other properties

PR #156, opened 2016-12-13
Do not save a playlist automatically if we are inside a project

vguzmanp:


VAS.Services/Controller/PlaylistController.cs
```diff
@@ -147,9 +147,8 @@ protected void LoadPlay (TimelineEvent play, Time seekTime, bool playing)
 			if (name != null) {
 				playlist = new Playlist { Name = name };
 				viewModel.Model.Add (playlist);
-				if (projectViewModel == null) {
-					Save (playlist, true);
-				}
+				Save (playlist, true);
+
```
ylatuya: extra blank line

VAS.Tests/Services/TestPlaylistController.cs
```diff
@@ -215,5 +223,185 @@ public void TestDeletePlaylist ()
...
+			};
+			App.Current.EventsBroker.Publish (ev);
+
+			// Assert
+			storageMock.Verify (s => s.Store<Playlist> (playlist, true), Times.Once ());
+			storageMock.Verify (s => s.Store<Playlist> (playlist2, true), Times.Once ());
+		}
+
+		[Test]
+		public void TestSavePlaylistMoveElementWithProject ()
```
vguzmanp: This is the only test affected by the bug, but I added all of them for completeness.

PR #157, opened 2016-12-16
RA-426

dfernandezramos:


vguzmanp:
The assemblyInfo file is changed on build. This changes will probably be lost when building again.

dfernandezramos:
But the version needs to be taken from there. When the app has been packaged it has been built so the version number will be the correct one.

PR #158, opened 2016-12-16
RA 228

vmartos:


VAS.Core/Resources.cs
```diff
@@ -29,7 +29,7 @@ public static Image LoadImage (string name)
 									Boolean.Parse (Environment.GetEnvironmentVariable (StyleConf.TEST)) :
 									false;
 			if (isTest) {
-				return new Image (10, 10);
+				return null;
```
aameijide: I think is better as before. Returning a null will make a crash if the test use some property from the image (as .Width, for instance). Also, when running the app this method usually will return an image (null will be only returned if file was not found, but it shouldn't be a common case).

VAS.Core/StyleConf.cs
```diff
@@ -291,10 +291,6 @@ public StyleConf ()
 
 		//Sizes
 
-		public int IconXLargeHeight { get; set; }
-
-		public int IconXLargeWidth { get; set; }
-
```
aameijide: Why are these properties deleted? They are defined in design guide

VAS.Services/Controller/PlaylistController.cs
```diff
@@ -106,7 +106,9 @@ public void SetViewModel (IViewModel viewModel)
 			}
 			this.viewModel = (PlaylistCollectionVM)(viewModel as dynamic);
 			// projectViewModel can be set to null...
-			this.projectViewModel = (ProjectVM)(viewModel as dynamic);
+			try {
```
aameijide: Is not possible with explicit cast to check `if (viewModel != null && (viewModel as dynamic) is ProjectVM)` instead of lei it crash?

VAS.Services/State/ScreenState.cs
```diff
@@ -19,6 +19,7 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using VAS.Core;
+using VAS.Core.Common;
```
aameijide: unused using

VAS.UI.Helpers.Gtk2/Bindings.cs
```diff
@@ -38,6 +47,10 @@ static public void Bind (this Button button, ICommand command, object parameter
 				command.CanExecuteChanged -= handler;
 			};
 			button.Clicked += (sender, e) => {
+				var radioButton = sender as RadioButton;
+				if (radioButton != null && radioButton.Active == false) {
+					return;
+				}
```
aameijide: Why we only execute the command if only is an inactive RadioButton?

VAS.Core/Common/StateController.cs
```diff
@@ -248,6 +257,26 @@ public void Register (string transition, Func<IScreenState> panel)
...
+		public void Register (string transition, Command command, Func<IScreenState> panel)
+		{
+			if (destination.Keys.Contains (transition)) {
+				if (!overwrittenTransitions.ContainsKey (transition)) {
+					overwrittenTransitions [transition] = new Stack<Func<IScreenState>> ();
+				}
+				overwrittenTransitions [transition].Push (panel);
+			}
+
+			destination [transition] = panel;
```
vmartos: The key is auto added if it doesn't exist. It crashes if you want to extract the key and doesn't exist. for example: panel = destination[transition]. but for setting new value if it exists the key it just overwrite the value if it doesn't exists it just creates the new key with the value.

Also I have refactor this method to call the Register(transition, panel) to not duplicate the code.

VAS.Core/Common/StateController.cs
```diff
@@ -248,6 +257,26 @@ public void Register (string transition, Func<IScreenState> panel)
...
+
+			transitionCommands.Add (transition, command);
```
vmartos: the previous code in if are only for overwritten transitions, the transition command can be overwritten by other tools, but cannot recover previous overwritten transition commands, because is not supported yet.

VAS.Core/StyleConf.cs
vmartos: I have talked with paolo, and now IconXLarge sizes are not needed anymore, because of the restriction to be squared size, and IconXLarge was 35x55, So now all IconXLarge has become IconLarge. Maybe the design guide should be updated.

VAS.UI.Helpers.Gtk2/Bindings.cs
vmartos: That's because radioButton executes the clickedEvent when they go from Inactive to Active and when they go from Active to Inactive, and we only want to execute the command when the RadioButton goes from Inactive to Active. That's why if radioButton.Active == false we don't execute the command.

VAS.Core/Common/StateController.cs
aameijide: 👍

VAS.Core/Filters/Predicate.cs
```diff
@@ -210,6 +210,14 @@ IEnumerator IEnumerable.GetEnumerator ()
 		}
 
 		#endregion
+
+		public void RaiseCollectionPropertyChange (object sender)
```
aoller: Changed

VAS.Core/MVVMC/NestedViewModel.cs
```diff
@@ -118,6 +131,42 @@ void HandleSelectionChanged (object sender, NotifyCollectionChangedEventArgs e)
...
+				RemoveNestedChildSelectionChanged (e.OldItems.OfType<INestedViewModel> ());
+				break;
+
+			case NotifyCollectionChangedAction.Reset:
+				RemoveNestedChildSelectionChanged (ViewModels.OfType<INestedViewModel> ());
+				break;
+
+			case NotifyCollectionChangedAction.Replace:
+				AddNestedChildSelectionChanged (e.NewItems.OfType<INestedViewModel> ());
+				RemoveNestedChildSelectionChanged (ViewModels.OfType<INestedViewModel> ());
```
aoller: It was a mistake. Changed.

VAS.UI.Helpers.Gtk2/Bindings.cs
aameijide: What happens if a button is clicked? It shouldn't execute the command?

aameijide: 👍

VAS.Core/ViewModel/TimelineVM.cs
```diff
@@ -168,21 +168,50 @@ protected virtual void HandleEventTypesCollectionChanged (object sender, NotifyC
 		{
 			eventTypeToTimeline.Clear ();
 			foreach (EventTypeTimelineVM timeline in ViewModels) {
-				eventTypeToTimeline [timeline.Model] = timeline;
+				eventTypeToTimeline [timeline.Model.Name] = timeline;
```
aameijide: Access through VM: `timeline.Name`

VAS.Core/MVVMC/NestedViewModel.cs
```diff
@@ -35,6 +36,9 @@ public NestedViewModel ()
 			ViewModels = new RangeObservableCollection<VMChilds> ();
 			Selection = new RangeObservableCollection<VMChilds> ();
 			Selection.CollectionChanged += HandleSelectionChanged;
+			if (ViewModels != null) {
```
aameijide: Still pending

aoller: Done

VAS.Tests/Core/ViewModel/TestTimelineVM.cs
```diff
@@ -102,5 +102,21 @@ public void TestAddTimelineEventWithEmptyEventTypes ()
...
+			ProjectVM projectVM = new ProjectVM { Model = project };
+			TimelineVM viewModel = new TimelineVM ();
+			viewModel.CreateEventTypeTimelines (projectVM.EventTypes);
+			viewModel.Model = project.Timeline;
+
+			viewModel.Clear ();
+
+			Assert.AreEqual (0, viewModel.ViewModels.Count);
+			Assert.AreEqual (0, viewModel.FullTimeline.ViewModels.Count);
+
```
aoller: Done

PR #159, opened 2016-12-19
Set the MachineGuid in Current_User/RiftAnalyst

vguzmanp:


VAS.Core/DeviceID.cs
```diff
@@ -126,11 +126,32 @@ public static Guid OSXDeviceID ()
...
 		{
+			Log.Debug ("Getting deviceID");
 			RegistryKey key = Microsoft.Win32.Registry.LocalMachine.OpenSubKey ("Software\\Microsoft\\Cryptography");
 			try {
-				return Guid.Parse (key.GetValue ("MachineGuid") as string);
+				var guid = key.GetValue ("MachineGuid") as string;
+				if (guid == null) {
+					Log.Debug ("LocalMachine MachineGuid is null");
+					RegistryKey userKey = Microsoft.Win32.Registry.CurrentUser.OpenSubKey ("Software", true)
+												   .CreateSubKey ("RiftAnalyst");
```
ylatuya: This should use the application name instead of RiftAnalyst

VAS.Core/DeviceID.cs
```diff
@@ -126,11 +126,32 @@ public static Guid OSXDeviceID ()
...
+		/// or from "HKCU\Software\RiftAnalyst\MachineGuid" if not. The latter will be created if needed.
+		/// </summary>
+		/// <returns>The device identifier.</returns>
 		public static Guid WindowsDeviceID ()
 		{
+			Log.Debug ("Getting deviceID");
 			RegistryKey key = Microsoft.Win32.Registry.LocalMachine.OpenSubKey ("Software\\Microsoft\\Cryptography");
 			try {
-				return Guid.Parse (key.GetValue ("MachineGuid") as string);
+				var guid = key.GetValue ("MachineGuid") as string;
```
ylatuya: I would use a single method that's known to be working everywhere, which is the 3rd one using CurrentUser

VAS.Core/DeviceID.cs
vguzmanp: D'oh! Fixed

VAS.Core/DeviceID.cs
vguzmanp: You mean using HKCU directly, right? I prefer using the global machineGuid if we can, as it's the one that actually identifies the device. The one per user is a workaround.

ylatuya: But this workaround is the vast majority of our users, probably around 80% of the Windows users based on the market share stats. The day we remove it with will see in our stats an increase of unique user of an 80%. So if we choose a solution, we need to stick to it

vguzmanp: But then we could just do this on every platform, we don't even need conditional platform code, we can write the newly generated guid to a file and read it.

The thing is that we are trying to use this to identify users, but this is the code for identifying devices. If we can identify a device, I think it's correct to actually identify it, even if it's the minority.

PR #160, opened 2016-12-19
Vas 41

ylatuya:


VAS.Core/ViewModel/TemplateViewModel.cs
```diff
@@ -15,18 +15,21 @@
...
 using VAS.Core.MVVMC;
+using VAS.Core.Store;
 
 namespace VAS.Core.ViewModel
 {
 	/// <summary>
 	/// Generic base class for <see cref="ITemplate"/> ViewModel.
 	/// </summary>
-	public abstract class TemplateViewModel<T> : ViewModelBase<T>, IViewModel<T> where T : ITemplate<T>
+	public abstract class TemplateViewModel<T> : ViewModelBase<T>, IViewModel<T> where T : ITemplate
```
vmartos: ViewModelBase<T> already implements IViewModel<T>

VAS.Tests/Core/ViewModel/TestTemplateVM.cs
```diff
@@ -59,5 +57,76 @@ public void TestEdited ()
...
+			Assert.AreEqual (3, viewModel.Count ());
+			Assert.IsInstanceOf (typeof (DummyPlayerVM), viewModel.First ());
+		}
+
+		[Test]
+		public void TestAddChild ()
+		{
+			var model = new DummyTeam {
+				Name = "dash",
+				Static = true,
```
vmartos: mm, if Static is True The Template shouldn't be editable! that means that we cannot add new players? Because this test are adding new players. Could be tested that we cannot edit a templated if Static is true?

ylatuya: If static is true the template can't be saved, but it can be edited. When the user tries to save the application ask the user if he wants to save it under a different name. In this unit tests Static does not have any impact, but I will remove it because it's confusing.

PR #163, opened 2016-12-20
Ra 327

vguzmanp:


aameijide:
Commit c50bf3632f1daf0a7b90e7d54fd3f08bf1c50eba was cherry-picked in RA-331 and now is in master, so when this branch is rebased should disappear. Also, review of that commit was made there.

VAS.Core/StyleConf.cs
```diff
@@ -357,6 +357,12 @@ public StyleConf ()
 
 		public int TitlesFontSize { get; set; }
 
+		public string TitlesFont {
+			get {
+				return string.Concat (TitlesFontFamily, " ", TitlesFontSlant, " ", TitlesFontSize);
+			}
+		}
+
```
aameijide: Remind: As this was introduced also in RA-331, so when rebasing commit e0929bd608328fc5ff27d2a4b6e6e7424fd354b7 just also disappear.

PR #166, opened 2016-12-28
Ra 391

ylatuya:


VAS.Core/Common/AsyncHelpers.cs
```diff
@@ -15,7 +15,10 @@
 //  along with this program; if not, write to the Free Software
 //  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 //
+using System.Linq;
 using System;
```
aameijide: order using

VAS.Core/Common/AsyncHelpers.cs
```diff
@@ -42,6 +45,31 @@ public static Task Return ()
 		{
 			return Task<bool>.FromResult (false);
 		}
+
+		/// <summary>
+		/// Parallelize tasks in partitions limiting the number of operations that are able to run in parallel
+		/// like <see cref="Parallel.ForEach"/> do but being an Asynchronous method.
+		/// https://blogs.msdn.microsoft.com/pfxteam/2012/03/05/implementing-a-simple-foreachasync-part-2/
```
aameijide: 👍  With that link it was easier to understand

PR #167, opened 2016-12-30
Adapt User statics service to code style

aameijide:


VAS.Services/UserStatisticsService.cs
```diff
@@ -106,7 +120,108 @@ public virtual bool Stop ()
...
+			return projectDictionary;
+		}
+
+		/// <summary>
+		/// Gets the timer list.
+		/// </summary>
+		/// <returns>The timer list.</returns>
+		public List<Tuple<string, int>> GetTimerList ()
+		{
+			return timerList;
```
dfernandezramos: As we have discussed, you can create properties instead of variables and define a public getter but a protected setter.

dfernandezramos:
👍

PR #169, opened 2017-01-03
Prevent that loaded template can be null

aameijide:
When returning from a canceled modal,loadedTemplate can be null.

### DIFF
--- a/VAS.Tests/Services/TestTemplatesController.cs
+++ b/VAS.Tests/Services/TestTemplatesController.cs
@@ -1,0 +1,77 @@
+﻿//
+//  Copyright (C) 2017 Fluendo S.A.
+//
+//  This program is free software; you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation; either version 2 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using System;
+using NUnit.Framework;
+
+namespace VAS.Tests.Services
+{
+	[TestFixture ()]
+	public class TestTemplatesController
+	{
+		DummyTemplatesController templatesController;
+
+		[SetUp]
+		public void SetUp ()
+		{
+			templatesController = new DummyTemplatesController ();
+		}
+
+		[Test ()]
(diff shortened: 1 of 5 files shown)


VAS.Services/Controller/TemplatesController.cs
```diff
@@ -429,8 +429,12 @@ protected bool IsChildSelection (IEnumerable<TViewModel> selectedViewModels)
 					return;
 				}
 			}
-			// Load the model
-			ViewModel.LoadedTemplate.Model = loadedTemplate;
+
+			// When returning from a canceled modal,loadedTemplate can be null
+			if (loadedTemplate != null) {
+				// Load the model
+				ViewModel.LoadedTemplate.Model = loadedTemplate;
```
ylatuya: loadedTemplate could be null, it's way of indicating that there is no template loaded.
The bug needs to be fixed somewhere else.

VAS.Tests/Utils.cs
```diff
@@ -123,6 +126,29 @@ public class DummyTeamVM : TemplateViewModel<DummyTeam, Utils.PlayerDummy, Dummy
...
+			handleSelectionChangedTimes = 0;
+			ViewModel = new TemplatesManagerViewModel<DummyTeam, DummyTeamVM, Utils.PlayerDummy, DummyPlayerVM> ();
+		}
+
+		protected override bool SaveValidations (DummyTeam model)
+		{
+			throw new NotImplementedException ();
+		}
+
+		protected override async void HandleSelectionChanged (object sender, PropertyChangedEventArgs e)
```
ylatuya: I don't think you need any of that, your test should check that with a template loaded in the controller, when we clear the selection the LoadTemplate.Model == null. A unit test should always try to test the public API not the internals, like it's doing here.

VAS.Tests/Services/TestTemplatesController.cs
```diff
@@ -0,0 +1,52 @@
...
+		{
+			// Arrange
+			DummyTeam team = new DummyTeam ();
+
+			// Action
+			templatesController.ViewModel.Selection.Add (null);
+
+			// Assert
+			Assert.AreEqual (1, templatesController.handleSelectionChangedTimes,
+							 $"HandleSelectionChanged was called {templatesController.handleSelectionChangedTimes} instead of 1");
```
aameijide: Extra error descriptions were added in order to easy identify which assert failed, as in this test we have 3.  Nunit description error doesn't show which value failed, it shows only "Expected: X but was: Y", so maybe instead of writing all the line ("Loaded template model is not null") we can only write the property name ("Loaded template model")

VAS.Tests/Utils.cs
```diff
@@ -123,6 +126,20 @@ public class DummyTeamVM : TemplateViewModel<DummyTeam, Utils.PlayerDummy, Dummy
 		}
 	}
 
+	public class DummyTemplatesController : TemplatesController<DummyTeam, DummyTeamVM, Utils.PlayerDummy, DummyPlayerVM>
+	{
+		public int handleSelectionChangedTimes;
```
ylatuya: this field should be removed

PR #170, opened 2017-01-03
Call LoadSurfaces only once per class using the static constructor

vguzmanp:


VAS.Drawing/CanvasObjects/Dashboard/LinkAnchorObject.cs
```diff
@@ -25,21 +25,25 @@
...
 
 		readonly int iconWidth;
 		readonly int iconHeight;
 		const int radius = 5;
 		double width;
 		double height;
 
+		static LinkAnchorObject ()
+		{
+			LoadSurfaces ();
```
ylatuya: I think it's more clear if the surfaces are loaded here, instead of calling another method

VAS.Drawing/CanvasObjects/Timeline/NeedleView.cs
```diff
@@ -28,9 +28,13 @@ public class NeedleView : CanvasObject, ICanvasSelectableObject
 		static ISurface needle;
 		static bool surfacesInitialized = false;
```
ylatuya: Not needed anymore

PR #175, opened 2017-01-05
RA-469: Add load image from icon in Misc

aameijide:


VAS.UI.Helpers.Gtk2/Misc.cs
```diff
@@ -235,11 +235,29 @@ public static Gdk.Pixbuf LoadIcon (string name, int size, IconLookupFlags flags
...
+		/// <returns>The icon as a Gdk.Image or missing image icon if not found.</returns>
+		/// <param name="name">Icon Name.</param>
+		/// <param name="width">Icon Width as a Gtk.IconSize.</param>
+		/// <param name="width">Icon Height as a Gtk.IconSize.</param>
+		/// <param name="flags">Lookup Flags like ForceSVG.</param>
+		public static Gtk.Image LoadImageIcon (string name, int width, int height, IconLookupFlags flags = IconLookupFlags.ForceSvg)
+		{
+			Gtk.Image image = new Gtk.Image ();
+			image.Pixbuf = Misc.LoadIcon (name, width);
+			image.WidthRequest = width;
```
vmartos: I don't think having a widthRequest and HeightRequest here is a good idea, by default all icon images has to be square sized: see https://fluendo.atlassian.net/wiki/display/LON/Graphic+resources%3A+icons+and+images#suk=ff808081520a3deb01521c0225e10002
So we need to only call Helpers.Misc.LoadIcon ("imagepath", StyleConf.Size)

VAS.UI.Helpers.Gtk2/Misc.cs
```diff
@@ -240,19 +240,18 @@ public static Gdk.Pixbuf LoadIcon (string name, Gtk.IconSize size, IconLookupFla
...
+		/// Loads the a image for a given name and size (icons must be squared)
 		/// </summary>
 		/// <returns>The icon as a Gdk.Image or missing image icon if not found.</returns>
 		/// <param name="name">Icon Name.</param>
-		/// <param name="width">Icon Width as a Gtk.IconSize.</param>
-		/// <param name="width">Icon Height as a Gtk.IconSize.</param>
+		/// <param name="size">Icon size as a int.</param>
 		/// <param name="flags">Lookup Flags like ForceSVG.</param>
-		public static Gtk.Image LoadImageIcon (string name, int width, int height, IconLookupFlags flags = IconLookupFlags.ForceSvg)
+		public static Gtk.Image LoadImageIcon (string name, int size, IconLookupFlags flags = IconLookupFlags.ForceSvg)
```
vmartos: this method is not needed now, it makes the same but just returning an image, In fact we want an image to be returned, we can let this method but i will do:

`public static Gtk.Image LoadImageIcon (string name, int size, IconLookupFlags flags = IconLookupFlags.ForceSvg)
{
     return new Gtk.Image(Misc.LoadIcon (name, size));
}`

aameijide: Yes, it seems logic, also, as we talked, is better to initialise directly than having a method.

aameijide:
Discarding this pull request. The changes that it should made doesn't have sense now.

PR #180, opened 2017-01-06
Ra 466

ylatuya:


VAS.Tests/LocalStorage.cs
```diff
@@ -0,0 +1,165 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using VAS.Core.Interfaces;
+using VAS.Core.Store;
+using VAS.Core.Filters;
+using VAS;
+
+namespace Tests
+{
+	public class LocalStorage : IStorage
```
aameijide: Document public methods

VAS.Tests/LocalStorage.cs
```diff
@@ -0,0 +1,165 @@
...
+		{
+			return localStorage.ContainsKey (t.ID);
+		}
+
+		public int Count<T> () where T : IStorable
+		{
+			return localStorage.Count;
+		}
+
+		public StorageInfo Info {
```
aameijide: Properties must go before methods

VAS.Tests/LocalStorage.cs
```diff
@@ -0,0 +1,165 @@
...
+		public bool Delete ()
+		{
+			return true;
+		}
+
+		#endregion
+	}
+
+
+	public class LocalDatabaseManager : IStorageManager
```
aameijide: I would split this in two different files

VAS.Tests/LocalStorage.cs
ylatuya: The file is moved from longomatch, It's a dummy implementation for unit tests, it does not need documentation

VAS.Tests/LocalStorage.cs
ylatuya: done

VAS.Tests/LocalStorage.cs
ylatuya: done

PR #181, opened 2017-01-10
RA-445

aoller:


VAS.DB/CouchbaseStorage.cs
```diff
@@ -218,6 +218,43 @@ public void Fill (IStorable storable)
 			}
 		}
 
+		public void StoreAll<T> (IEnumerable<T> tt, bool forceUpdate = false) where T : IStorable
```
ylatuya: This function is a copy paste of the Store function.
This function needs to be added to the IStorage API.
I would not change the name of the function and keep using Store<T> (IEnumerable<T> items, bool forceUpdate = false)
Use a meaningful name for the vairable tt.

ylatuya: The new function is missing unit tests

VAS.Core/Serialization/Serializer.cs
```diff
@@ -162,7 +162,7 @@ public T LoadSafe<T> (string filepath)
 				settings.ObjectCreationHandling = ObjectCreationHandling.Replace;
 				settings.Converters.Add (new VersionConverter ());
 				settings.Converters.Add (new VASConverter (true));
-				//settings.ReferenceResolver = new IdReferenceResolver ();
+				settings.ReferenceLoopHandling = ReferenceLoopHandling.Serialize;
```
ylatuya: Why is this needed? Which new object introduces loop references?

aoller: SummonerImage serialization requires this flag. In fact I didn't understant why it's needed. All uint tests works fine with this flag set.

VAS.DB/CouchbaseStorage.cs
aoller: Changed

VAS.Core/Serialization/Serializer.cs
ylatuya: There are no loop references in this Storable, there is probably probably something wrong somewhere else because it shouldn't be needed at all.

VAS.DB/CouchbaseStorage.cs
```diff
@@ -218,6 +218,43 @@ public void Fill (IStorable storable)
 			}
 		}
 
+		public void Store<T> (IEnumerable<T> storableEnumerable, bool forceUpdate = false) where T : IStorable
```
ylatuya: This function is a c&p of the Store one with one element

VAS.Tests/DB/TestStorage.cs
```diff
@@ -467,5 +467,31 @@ public void TestPlaylistDelete ()
...
+			Assert.IsNotNull (p1o);
+			Assert.AreEqual (p1.ID, p1o.ID);
+			Assert.AreEqual (p1.Name, p1o.Name);
+			Assert.IsNotNull (p2o);
+			Assert.AreEqual (p2.ID, p2o.ID);
+			Assert.AreEqual (p2.Name, p2o.Name);
+		}
+	}
+
+	class PlayerTest : Player
```
ylatuya: There is already a PlayerDummy

VAS.DB/CouchbaseStorage.cs
aoller: Refactored

VAS.Tests/DB/TestStorage.cs
aoller: Changed

VAS.Tests/DB/TestStorage.cs
```diff
@@ -467,5 +468,27 @@ public void TestPlaylistDelete ()
 			// Assert
 			Assert.AreEqual (1, db.DocumentCount);
 		}
+
+		[Test ()]
+		public void TestStorableEnumerable ()
+		{
+			// Arrange
+			PlayerDummy p1 = new Utils.PlayerDummy { Name = "P1" };
```
ylatuya: Using this the PlayerDummy defined in the using?

aoller: Changed

PR #182, opened 2017-01-12
RA-493

aoller:


VAS.Core/Store/StorableBase.cs
```diff
@@ -98,6 +100,16 @@ public StorableBase ()
 
 		#endregion
 
+		#region CreationDate
+
+		[PropertyPreload]
+		public DateTime CreationDate {
```
aameijide: Document this property

VAS.Core/Store/StorableBase.cs
```diff
@@ -98,6 +100,16 @@ public StorableBase ()
 
 		#endregion
 
+		#region CreationDate
```
aameijide: As a suggestion, I think this region isn't necessary, there's only a property inside and that doesn't come from a interface.

PR #184, opened 2017-01-12
Ra 397 bis

aameijide:


VAS.Services/VideoPlayerController.cs
```diff
@@ -1196,6 +1193,7 @@ protected virtual void Reset ()
 			loadedSegment.Start = new Time (-1);
 			loadedSegment.Stop = new Time (int.MaxValue);
 			loadedEvent = null;
+			UpdatePlayingState (true);
```
ylatuya: false instead of true. There must be a unit test for it

VAS.Core/MVVMC/ViewModelBase.cs
```diff
@@ -35,6 +35,11 @@ public class ViewModelBase<T> : BindableBase, IViewModel<T> where T : INotifyPro
 			set;
 			get;
 		}
+
+		protected override void ForwardPropertyChanged (object sender, PropertyChangedEventArgs e)
+		{
+			base.ForwardPropertyChanged (this, e);
```
ylatuya: Unit tests ?¿

PR #188, opened 2017-01-16
Ra 519

vguzmanp:


HockeySDK-Windows/Src/Kit.Core45/Channels/Persistence/Sender.cs
```diff
@@ -4,339 +4,282 @@
...
-
-        /// <summary>
-        /// Disposes the managed objects.
-        /// </summary>
-        public void Dispose()
-        {
-            if (Interlocked.Increment(ref this.disposeCount) == 1)
-            {
-                this.StopAsync().ConfigureAwait(false).GetAwaiter().GetResult();
+	using System;
```
ylatuya: This time it makes sense splitting formatting and changes because it's an external project and it easier to understand what what changes in case one day we need to sync again with upstream

VAS.Core/KPI/KpiService.cs
```diff
@@ -44,7 +44,8 @@ public void TrackEvent (string eventName, IDictionary<string, string> properties
 		public void TrackException (Exception ex, IDictionary<string, string> properties)
 		{
 #if !DEBUG
-			HockeyClient.Current.TrackException (ex, properties);
+			HockeyClient.Current.HandleException (ex);
+			HockeyClient.Current.SendCrashesAsync ().ConfigureAwait (false);
```
ylatuya: If this is needed because HA's Flush isn't doint it? Shouldn't we do it in  our Flush implementation ?

VAS.Core/KPI/HockeyClientVASExtensions.cs
```diff
@@ -36,7 +36,7 @@ public static class HockeyClientVASExtensions
 		/// <param name="identifier">Identfier.</param>
 		/// <param name="keepRunningAfterException">Keep running after exception.</param>
 		/// <returns>Instance object.</returns>
-		public static IHockeyClientConfigurable Configure (this IHockeyClient @this, string identifier)
+		public async static Task<IHockeyClientConfigurable> Configure (this IHockeyClient @this, string identifier)
```
ylatuya: Now that your are on it, change the parameter @this with something else. There is no reason to use a reserved word for it :)

HockeySDK-Windows/Src/Kit.Core45/Channels/Persistence/Sender.cs
vguzmanp: Done

VAS.Core/KPI/KpiService.cs
vguzmanp: Hmm... the main change here is that instead of calling `TrackException` it's calling `HandleException`. The first one sends the event as a `exception` object to Insights, the second sends a crash to HockeyApp.

The HandleException already does a Flush internally I think, but it only writes it to the storage. The SendCrashes is the one that actually sends it.

VAS.Core/KPI/HockeyClientVASExtensions.cs
vguzmanp: I thought about this, but I preferred to leave it like this in case we need to sync with upstream.

VAS.Core/KPI/KpiService.cs
ylatuya: I am talking about HockeyClient.Current.SendCrashesAsync ().ConfigureAwait (false);
SendCrashes should only be forced when Flushing.

VAS.Core/KPI/HockeyClientVASExtensions.cs
ylatuya: There is nothing to sync  here, that's our code not HA's HockeySDK.Net45 we imported

VAS.Core/Common/Utils.cs
```diff
@@ -87,6 +87,18 @@ public static string SanitizePath (string path, params char[] replaceChars)
 			}
 		}
 
+		/// <summary>
+		///     Returns the OS version number for this computer.
+		/// </summary>
+		public static string GetOSVersion ()
```
ylatuya: I'd make it a property to have the same API: Utils.OS Utils.OSVersion

ylatuya: And cache the result, like with the device id

VAS.Core/Common/Utils.cs
```diff
@@ -130,6 +142,53 @@ public static Stream GetEmbeddedResourceFileStream (string resourceId)
...
+				Log.Warning ("Failed getting OS version. Falling back to Environment");
+				version = GetOSVersionFallback ();
+			}
+
+			return version;
+		}
+
+		static string GetOSVersionFallback ()
+		{
+			return string.Format ("{0} {1} {2}", OS, Environment.OSVersion.Version, Environment.OSVersion.ServicePack);
```
ylatuya: you can use the coolest C# 6 feature ever:
return $"{OS} {Environment.OSVersion.Version} {Environment.OSVersion.ServicePack}"

PR #193, opened 2017-01-19
RA-340

vmartos:


vmartos:
Yes this affects longomatch, and should be changed in the future. But the changes shouldn't be that much

VAS.Services/Controller/TaggingController.cs
```diff
@@ -130,124 +120,39 @@ protected void HandleClickedPCardEvent (ClickedPCardEvent e)
...
-			play.Start.MSeconds = Math.Max (0, play.Start.MSeconds);
-			if (project.ProjectType == ProjectType.FileProject) {
-				play.Stop.MSeconds = Math.Min (project.FileSet.Duration.MSeconds, play.Stop.MSeconds);
-				play.CamerasLayout = videoPlayer.CamerasLayout;
-				play.CamerasConfig = new ObservableCollection<CameraConfig> (videoPlayer.CamerasConfig);
-			} else {
-				play.CamerasLayout = null;
-				play.CamerasConfig = new ObservableCollection<CameraConfig> { new CameraConfig (0) };
-			}
+			if (project == null)
```
dfernandezramos: Use braces

VAS.Services/Controller/TaggingController.cs
```diff
@@ -128,39 +130,132 @@ protected void HandleClickedPCardEvent (ClickedPCardEvent e)
...
+			Log.Debug (String.Format ("New play created start:{0} stop:{1} category:{2}",
+				play.Start.ToMSecondsString (), play.Stop.ToMSecondsString (),
+				play.EventType.Name));
+			project.Model.Timeline.Add (play);
+			AddNewPlay (play);
+
 			Reset ();
 		}
 
+		protected abstract TimelineEvent CreateTimelineEvent (EventType type, Time start, Time stop,
```
vmartos: It was github!

VAS.Services/Controller/TaggingController.cs
```diff
@@ -86,8 +84,12 @@ public void Stop ()
 		/// <param name="viewModel">View model.</param>
 		public void SetViewModel (IViewModel viewModel)
 		{
-			ViewModel = (ProjectVM)(viewModel as dynamic);
+			project = (ProjectVM)(viewModel as dynamic);
 			VideoPlayer = (VideoPlayerVM)(viewModel as dynamic);
+			try {
+				capturer = (ICapturerBin)(viewModel as dynamic);
+			} catch {
```
vmartos: this was commented in an outdated diff. I refactor TaggingController and removed almost all code that i was previously added. Now the TaggingController doesn't needs a capturer

VAS.Core/ViewModel/DashboardButtonVM.cs
```diff
@@ -436,5 +438,54 @@ public class TimerButtonVM : DashboardButtonVM
...
+				return currentTime;
+			}
+
+			set {
+				if (value != null && currentNode != null) {
+					if (currentTime < currentNode.Start) {
+						TimerTime = null;
+						currentNode = null;
+					}
+					if (currentTime != null &&
```
vmartos: "When you perform comparisons with nullable types, if the value of one of the nullable types is null and the other is not, all comparisons evaluate to false except for != (not equal)."
https://msdn.microsoft.com/en-us/library/2cf62fcy.aspx
So comparing (null < Time) doesn't crash, but for this reason also it doesn't make sense to compare if currentTime is null so I added the currentTime != null in the first if

dfernandezramos:
👍

PR #194, opened 2017-01-20
Vas 44

dfernandezramos:


VAS.Core/Common/Stopwatch.cs
```diff
@@ -0,0 +1,89 @@
...
+using VAS.Core.Interfaces;
+
+namespace VAS.Core.Common
+{
+	/// <summary>
+	/// VAS Stopwatch class that inherits from IStopwatch
+	/// </summary>
+	public class Stopwatch : IStopwatch
+	{
+		System.Diagnostics.Stopwatch stopwatch;
```
vguzmanp: Add using

VAS.Core/Common/Stopwatch.cs
```diff
@@ -0,0 +1,89 @@
...
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using System;
+using VAS.Core.Interfaces;
+
+namespace VAS.Core.Common
+{
+	/// <summary>
+	/// VAS Stopwatch class that inherits from IStopwatch
```
vguzmanp: My comment here would be something like "Wrapper for System.Diagnostics.Stopwatch"

VAS.Services/UserStatisticsService.cs
```diff
@@ -162,6 +168,15 @@ public UserStatisticsService ()
 
 		#region IService implementation
 
+		public UserStatisticsService ()
```
vguzmanp: Constructors should go in the top of the class, after fields

VAS.Services/UserStatisticsService.cs
```diff
@@ -248,9 +263,11 @@ void LoadSessionProjectValues (Guid projectId)
 		/// </summary>
 		void SaveTimer ()
 		{
-			stateTimer.Stop ();
-			TimerList.Add (new Tuple<string, long> (currentState, stateTimer.ElapsedTicks));
-			stateTimer.Reset ();
+			StateTimer.Stop ();
+			if (StateTimer.ElapsedMilliseconds >= 1000) {
```
vguzmanp: Why is this? timers shorter than 1 second are not stored?

PR #198, opened 2017-01-31
Add cancellation task support in ForEachAsync helper

aoller:


VAS.Core/Common/AsyncHelpers.cs
```diff
@@ -55,18 +56,32 @@ public static Task Return ()
...
+		public static Task ForEachAsync<T> (this IEnumerable<T> source, int dop, Func<T, Task> body, CancellationToken? ct = null)
 		{
 			return Task.WhenAll (
 				Partitioner.Create (source).GetPartitions (dop).Select (partition => {
-					return Task.Run (async () => {
-						using (partition) {
-							while (partition.MoveNext ()) {
-								await body (partition.Current);
+					if (!ct.HasValue) {
+						return Task.Run (async () => {
```
ylatuya: Why do you keep this path at all instaed of using the same with out without cancellation token? If there is no cancellation token create a default one.

VAS.Core/Common/AsyncHelpers.cs
```diff
@@ -55,18 +56,32 @@ public static Task Return ()
...
 							}
-						}
-					});
+						});
+					} else {
+						return Task.Factory.StartNew (async () => {
+							using (partition) {
+								while (partition.MoveNext ()) {
+									await body (partition.Current);
+									if (ct.GetValueOrDefault ().IsCancellationRequested) {
```
ylatuya: Use https://msdn.microsoft.com/en-us/library/system.threading.cancellationtoken.throwifcancellationrequested(v=vs.110).aspx

VAS.Core/Common/AsyncHelpers.cs
aoller: Changed

VAS.Core/Common/AsyncHelpers.cs
aoller: Changed

PR #199, opened 2017-02-01
Fix dispose implementations

vguzmanp:
This started from RA-446, but something in these disposes is wrong and needs to be fixed, as it's crashing the application.

### DIFF
--- a/VAS.UI.Multimedia.Gtk2/UI/VideoPlayerView.cs
+++ b/VAS.UI.Multimedia.Gtk2/UI/VideoPlayerView.cs
@@ -119,6 +119,37 @@ namespace VAS.UI
 
 		}
 
+		public override void Dispose ()
+		{
+			Dispose (true);
+			base.Dispose ();
+		}
+
+		protected virtual void Dispose (bool disposing)
+		{
+			if (Disposed) {
+				return;
+			}
+			if (disposing) {
+				Destroy ();
+			}
+			Disposed = true;
+		}
+
+		protected override void OnDestroyed ()
+		{
+			Log.Verbose ($"Destroying {GetType ()}");
+
+			playerVM = null;
+			blackboard.Dispose ();
+
+			base.OnDestroyed ();
+
+			Disposed = true;
+		}
+
+		protected bool Disposed { get; private set; } = false;
+
 		#endregion
 
 		protected override void OnRealized ()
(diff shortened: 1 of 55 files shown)


VAS.Tests/Services/TestControllerBase.cs
```diff
@@ -0,0 +1,264 @@
...
+			controller.managedDisposeCalled += ManagedDisposeCalled;
+			controller.unmanagedDisposeCalled += UnmanagedDisposeCalled;
+		}
+
+		[TearDown]
+		public void TearDown ()
+		{
+			if (controller != null) {
+				controller.managedDisposeCalled -= ManagedDisposeCalled;
+				controller.unmanagedDisposeCalled -= UnmanagedDisposeCalled;
```
dfernandezramos: Maybe stop the controller if it's started?

VAS.Tests/Services/TestControllerBase.cs
```diff
@@ -0,0 +1,264 @@
...
+			controller.Stop ();
+
+			// Act & Assert
+			Assert.Throws<InvalidOperationException> (controller.Stop);
+		}
+
+		[Test]
+		public void TestStopWithoutStart ()
+		{
+			// Act & Assert
```
dfernandezramos: Maybe do the same as you have done in TestDisposeStopped:
`// Arrange & Assume`
`Assert.IsFalse (controller.Started);`

VAS.Tests/Services/TestControllerBase.cs
```diff
@@ -0,0 +1,264 @@
...
+		}
+
+		[Test]
+		public void TestDisposeGC ()
+		{
+			// Arrange
+			GC.Collect ();
+			GC.WaitForPendingFinalizers ();
+			managedDisposeCalled = false;
+			unmanagedDisposeCalled = false;
```
dfernandezramos: You could Assert that managedDisposeCalled and unmanagedDisposeCalled are true. (Same in TestPublishBeforeGC() and TestPublishAfterGC())

VAS.UI.Multimedia.Gtk2/UI/VideoPlayerView.cs
```diff
@@ -521,14 +509,15 @@ protected virtual void HandleJumpValueChanged (object sender, EventArgs e)
 
 		protected virtual void HandleReady (object sender, EventArgs e)
 		{
-			viewPortsBackup = new List<IViewPort> { videowindow };
+			var viewPortsBackup = new List<IViewPort> { videowindow };
 			playerVM.ViewPorts = viewPortsBackup;
```
dfernandezramos: Isn't it better to just write `playerVM.ViewPorts = new List<IViewPort> { videowindow };`?

dfernandezramos:
👍

PR #204, opened 2017-02-07
Ra 446 without disposes

vguzmanp:
Some Dispose is wrong and it's crashing the app in a bad way. This is the minimum needed to fix RA-446


dfernandezramos:
resolve conflicts and 👍

PR #206, opened 2017-02-10
Make TreeViewBase subscribe to the correct overriden collection

vguzmanp:


VAS.Core/MVVMC/NestedViewModel.cs
```diff
@@ -123,7 +123,7 @@ public INotifyCollectionChanged GetNotifyCollection ()
 		/// Gets the Interface INotifyCollectionChanged of the Selection collection
 		/// </summary>
 		/// <returns>The Collection as a INotifyCollectionChanged</returns>
-		public INotifyCollectionChanged GetSelectionNotifyCollection ()
+		public virtual INotifyCollectionChanged GetSelectionNotifyCollection ()
```
aameijide: No need to make it virtual

vguzmanp: I made this virtual out of consistency. I understand that if the `GetNotify` is virtual, the `GetSelectionNotify` should be too.

aameijide: ok

PR #208, opened 2017-02-13
RA-219: Ads

aameijide:


VAS.UI.Helpers.Gtk2/ButtonHelper.cs
```diff
@@ -90,6 +90,16 @@ public static void ApplyStyleTab (this Button button, Command command = null)
...
 		}
 
+		public static void ApplyStyleRemove (this Button button, Command command = null)
+		{
+			if (command != null) {
+				button.Bind (command);
+			}
+			button.ApplyStyle (StyleConf.ButtonRemove, App.Current.Style.ButtonRemoveWidth);
+			button.HeightRequest = App.Current.Style.ButtonRemoveHeight;
+			button.SetAlignment (0.0f, 0.5f);
```
vmartos: Could it be possible to set HeightRequest and SetAlignment in the widget that is creating the buttons? this method is only for applying style

PR #211, opened 2017-02-15
Create a StartProcess util method

vguzmanp:


VAS.Core/Common/Utils.cs
```diff
@@ -168,6 +170,37 @@ public static Stream GetEmbeddedResourceFileStream (string resourceId)
...
+			Process proc = new Process ();
+			proc.EnableRaisingEvents = false;
+
+			proc.StartInfo.WorkingDirectory = Path.GetFullPath (Path.Combine (App.Current.baseDirectory, "bin"));
+			foreach (string variable in environmentVariables.Keys) {
+				proc.StartInfo.EnvironmentVariables [variable] = environmentVariables [variable];
+			}
+
+			if (System.Environment.OSVersion.Platform == PlatformID.Win32NT) {
+				proc.StartInfo.FileName = Path.Combine (App.Current.baseDirectory, "bin", $"{filename}.{windowsExtension}");
```
ylatuya: the first part of the path can be cached or taken from proc.StartInfo.WorkingDirectory.

ylatuya: Extension must be ".exe" instead of "exe". Why do you need it anyway ? I wouldn't use platform specific arguments.

vguzmanp: I was going to just add ".exe" to the filename, but I thought that someone could need to run a different thing.
Rechecking the docs, using `UseShellExecute = false`we can only run exes, so we don't need it.

vguzmanp: fixed

PR #213, opened 2017-02-16
Ra 497

vguzmanp:


VAS.Services/Controller/PlaylistController.cs
```diff
@@ -47,6 +46,16 @@ public class PlaylistController : ControllerBase
 			set;
 		}
 
+		/// <summary>
+		/// Gets or sets the project view model.
+		/// Set this ViewModel to make <see cref="PlaylistController"/> save the playlist to a project.
+		/// </summary>
+		/// <value>The project view model.</value>
+		protected ProjectVM ProjectViewModel {
```
vguzmanp: No, that's what the commit is about. I removed it from VAS, but the VAS code can still work with it. If LongoMatch needs it, it just needs to set it and everything will work as it did before.

VAS.Services/Controller/PlaylistController.cs
```diff
@@ -232,10 +237,10 @@ void HandleRenderPlaylist (RenderPlaylistEvent e)
 					}
 					questioned = true;
 				}
-				if (projectViewModel == null) {
+				if (ProjectViewModel == null) {
 					App.Current.DatabaseManager.ActiveDB.Delete (playlistVM.Model);
 				}
-				playlistViewModel.ViewModels.Remove (playlistVM);
+				playlistViewModel.Model.Remove (playlistVM.Model);
```
vguzmanp: This is the same that happened in the `ProjectController`. Removing from the `LimitationViewModel` doesn't update the LimitedViewModels correctly, while doing it from the Model does.
And I find it right, as this way we don't need to trust the VM to sync everything: all VM should sync correctly with changes to the Model.

VAS.Core/ViewModel/PlaylistCollectionVM.cs
```diff
@@ -45,10 +46,28 @@ public PlaylistCollectionVM ()
...
+		/// <value>The limitation.</value>
+		public override LicenseLimitationVM Limitation {
+			set {
+				if (Limitation != null) {
+					Limitation.PropertyChanged -= HandleLimitationChanged;
+				}
+				base.Limitation = value;
+				if (Limitation != null) {
+					Limitation.PropertyChanged += HandleLimitationChanged;
+					NewCommand.EmitCanExecuteChanged ();
```
vguzmanp: I had doubts here, as `EmitCanExecute` is such a basic thing, I didn't think it was needed to call the other method.

VAS.Core/ViewModel/PlaylistCollectionVM.cs
```diff
@@ -45,10 +46,28 @@ public PlaylistCollectionVM ()
...
+
+		/// <summary>
+		/// Sets the limitation view model.
+		/// </summary>
+		/// <value>The limitation.</value>
+		public override LicenseLimitationVM Limitation {
+			set {
+				if (Limitation != null) {
+					Limitation.PropertyChanged -= HandleLimitationChanged;
+				}
```
vguzmanp: I don't follow you here

VAS.UI.Gtk2/UI/Menu/MenuHelpers.cs
```diff
@@ -82,6 +83,9 @@ static public void FillAddToPlaylistMenu (MenuItem addToPlaylistMenu, IEnumerabl
 				};
 			}
 			item = new MenuItem (Catalog.GetString ("Create new playlist..."));
+			//FIXME: When using a VM here and limitations on license are controlled, check correctly if limitation is enabled.
```
aameijide: To delete

PR #214, opened 2017-02-16
RA-545: TimeNode returns a default start/end time if Model is null

aameijide:


VAS.Core/ViewModel/TimeNodeVM.cs
```diff
@@ -50,7 +50,7 @@ public TimeNodeVM ()
 		/// </summary>
 		public Time Start {
 			get {
-				return Model.Start;
+				return Model?.Start ?? new Time (0);
```
ylatuya: I would handle that diferently. In the MediaFileSetVM if model == null, VisibleRegion.Model = new TimeNode (0, 0)

ylatuya:
Also, missing unit tests.

aameijide:
About missing test, I don't think that a unit test to check if is null return (or sets in your approach) 0 is worthy.

ylatuya:
As I commented in the code, this is not the correct fix, it's in MediaFileSetVM. It always make sense to do write test, specially if, like in this case, it was causing a bug.

PR #215, opened 2017-02-17
Ra 617: Fix error handling in create encoder for render videos

vmartos:


VAS.Services/RenderingJobsController.cs
```diff
@@ -366,7 +366,7 @@ void OnError (object sender, string message)
 			App.Current.Dialogs.ErrorMessage (Catalog.GetString ("An error has occurred in the video editor.")
 			+ Catalog.GetString ("Please, try again."));
 			ViewModel.CurrentJob.State = JobState.Error;
-			StartNextJob ();
+			CancelCurrentJob (true);
```
ylatuya: CancelCurrentJobs sets the state to Cancelled instaed of Error. You might just want to CleanVideoEditor (true); before StartNextJob if there is an error

ylatuya: Unit tests missing

PR #216, opened 2017-02-20
Do not check selection if it's already disposed

vmartos:


VAS.Drawing/CanvasObjects/Timeline/TimeNodeView.cs
```diff
@@ -215,6 +215,10 @@ protected override void DisposeManagedResources ()
 		/// <param name="inMotion">If set to <c>true</c> in motion.</param>
 		public virtual Selection GetSelection (Point point, double precision, bool inMotion = false)
 		{
+			if (Disposed) {
+				return null;
+			}
```
vguzmanp: Does this happen anywhere else? If this is called it means that somewhere we are calling an already disposed object, so it's an error in the caller.

VAS.Drawing/Widgets/PlaysTimeline.cs
```diff
@@ -142,6 +142,17 @@ public virtual void SetViewModel (object viewModel)
...
 			base.ClearObjects ();
+			foreach (var vm in viewModelToView.Keys) {
+				var eventTypeVM = vm as EventTypeTimelineVM;
+				if (eventTypeVM != null) {
+					eventTypeVM.ViewModels.CollectionChanged -= HandleChildsVMCollectionChanged;
+					continue;
+				}
+				var timerVM = vm as TimerVM;
+				if (timerVM != null) {
+					timerVM.ViewModels.CollectionChanged -= HandleChildsVMCollectionChanged;
```
vguzmanp: Can't you cast it to CollectionViewModel, so that you can always call `.ViewModels`?

vguzmanp:
Squash & :+1:

PR #219, opened 2017-02-21
RA-642: Change StyleConf font to be setted by a derived class

aameijide:


VAS.Core/StyleConf.cs
```diff
@@ -361,6 +360,11 @@ public StyleConf ()
 
 		//Fonts
 
+		public virtual string Font {
+			get { return font; }
+			protected set { font = value; }
```
vmartos: you maybe can do:
public string Font {
    get;
    protected set;
} = "Ubuntu";
or in the constructor set Font = "Ubuntu", but as you prefer!
but I think that the virtual can be removed since it is not overriden by RA

PR #220, opened 2017-02-22
Close player on Unready and open Fileset on Ready

vmartos:


VAS.Services/VideoPlayerController.cs
```diff
@@ -364,12 +364,12 @@ public virtual void Ready (bool ready)
 					delayedOpen = null;
 				} else if (FileSet == null || !FileSet.Any ()) {
 					ShowMessageInViewPorts (Catalog.GetString ("No video available"), true);
+				} else {
+					Open (FileSet, false);
```
vguzmanp: Log.Debug this

VAS.Services/VideoPlayerController.cs
```diff
@@ -364,12 +364,12 @@ public virtual void Ready (bool ready)
...
 				}
 			} else {
-				Log.Debug ("Player unready");
-				if (Playing) {
-					Stop (false);
-				}
+				Log.Debug ("Player unready, closing player");
+				player?.Close ();
```
vguzmanp: if player is null here, it means that it's already disposed, so there is a problem somewhere else

PR #221, opened 2017-02-22
Allow having null in the limitationVM

vguzmanp:


vguzmanp:
TODO: Add unit tests for the command-limitation interactions

ylatuya:
This one is not fixed https://github.com/fluendo/VAS/blob/longomatch-fix/VAS.UI.Gtk2/UI/Menu/MenuHelpers.cs#L86
Also, it's named Presentations... why ?

vguzmanp:
The thing about Presentations is that we had a mixture of names (Playlist/Presentations) at that moment. I think the final decision was to use the name "Presentations", and then everyone started calling them playlists 😄

PR #224, opened 2017-02-23
Longomatch fix

ylatuya:


VAS.Core/ViewModel/PlayerTimelineVM.cs
```diff
@@ -0,0 +1,60 @@
...
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using System.Linq;
+using VAS.Core.Interfaces.GUI;
+using VAS.Core.Interfaces.MVVMC;
+using VAS.Core.MVVMC;
+using VAS.Core.Store;
+namespace VAS.Core.ViewModel
+{
+	public class PlayerTimelineVM : NestedViewModel<TimelineEventVM>, IViewModel<Player>
```
vguzmanp: Missing documentation

VAS.Core/ViewModel/PlayerTimelineVM.cs
```diff
@@ -0,0 +1,60 @@
...
+	{
+		public PlayerTimelineVM ()
+		{
+		}
```
vguzmanp: Redundant constructor

VAS.Core/ViewModel/PlayerTimelineVM.cs
```diff
@@ -0,0 +1,60 @@
...
+		public PlayerVM Player {
+			get;
+			protected set;
+		}
+
+		/// <summary>
+		/// Gets the total visible events inside this EventType
+		/// </summary>
+		/// <value>The visible events.</value>
+		public int VisibleEvents {
```
vguzmanp: We should have this in an interface / base class for all the timelines, as both eventTypeTimeline & PlayerTimeline define it.

VAS.Core/ViewModel/TimelineVM.cs
```diff
@@ -269,11 +296,35 @@ void AddTimelineEventVM (TimelineEventVM viewModel)
 				EventTypesTimeline.ViewModels.Add (new EventTypeTimelineVM { Model = viewModel.Model.EventType });
 			}
 			eventTypeToTimeline [viewModel.Model.EventType.Name].ViewModels.Add (viewModel);
+			UpdatePlayersTimeline (viewModel);
+		}
+
+		void UpdatePlayersTimeline (TimelineEventVM viewModel)
```
vguzmanp: This should be called `AddToPlayers` or something like that.

VAS.Core/ViewModel/TimelineVM.cs
```diff
@@ -269,11 +296,35 @@ void AddTimelineEventVM (TimelineEventVM viewModel)
...
+		{
+			foreach (Player player in viewModel.Model.Players) {
+				playerToTimeline [player].ViewModels.Add (viewModel);
```
vguzmanp: Why is this one adding to all the players? Shouldn't it add to the players tagged in the event?

VAS.Core/ViewModel/TimelineVM.cs
vguzmanp: And the parameter should be `event` or something like that

VAS.Tests/Core/ViewModel/TestTimelineVM.cs
```diff
@@ -134,5 +138,149 @@ public void TestRenameEventType ()
...
+			PlayerVM player1 = homeTeamVM.ViewModels [0];
+			PlayerVM player2 = awayTeamVM.ViewModels [0];
+			AddTimelineEventWithPlayers (timeline, player1, player2);
+
+			CheckPlayerEvents (timeline, player1, 1);
+			CheckPlayerEvents (timeline, player2, 1);
+		}
+
+		[Test]
+		public void TestAddNewEvent_WithPlayersFromOneTeam ()
```
vguzmanp: The names of this test and the one before are swapped

VAS.Tests/Core/ViewModel/TestTimelineVM.cs
```diff
@@ -134,5 +138,149 @@ public void TestRenameEventType ()
...
+			TimelineVM timeline;
+
+			CreateTimelineWithTeams (out timeline, out homeTeamVM, out awayTeamVM);
+
+			PlayerVM player1 = homeTeamVM.ViewModels [0];
+			PlayerVM player2 = homeTeamVM.ViewModels [1];
+			AddTimelineEventWithPlayers (timeline, player1, player2);
+
+			CheckPlayerEvents (timeline, player1, 1);
+			CheckPlayerEvents (timeline, player2, 1);
```
vguzmanp: Is this testing anywhere that there's only one TeamTimeline?

vguzmanp: This should also check the number of player/team timelines after adding the event

VAS.Drawing/CanvasObjects/Dashboard/AnalysisEventButtonView.cs
```diff
@@ -128,10 +124,16 @@ protected override void DisposeManagedResources ()
 			}
 		}
 
+		new AnalysisEventButton Button {
+			get {
+				return base.Button as AnalysisEventButton;
```
vguzmanp: return `ViewModel.Model`

VAS.Drawing/Widgets/DashboardCanvas.cs
```diff
@@ -438,24 +423,68 @@ protected virtual void HandleTaggerClickedEvent (ICanvasObject co)
...
 			}
-
-			if (e == null || e.PropertyName == "Mode") {
+			if (ViewModel.NeedsSync (e, nameof (ViewModel.Mode))) {
 				HandleModeChanged ();
-			} else if (e == null || e.PropertyName == "ShowLinks") {
+			} else if (ViewModel.NeedsSync (e, nameof (ViewModel.ShowLinks))) {
 				HandleShowLinksChanged ();
-			} else if (e == null || e.PropertyName == "FitMode") {
+			} else if (ViewModel.NeedsSync (e, nameof (ViewModel.FitMode))) {
```
vguzmanp: the elses will not be executed on Sync

VAS.Core/MVVMC/Binding.cs
```diff
@@ -0,0 +1,70 @@
...
+		bool binded;
+
+		public IViewModel ViewModel {
+			get {
+				return viewModel;
+			}
+			set {
+				viewModel = value;
+				BindViewModel ();
+				if (!binded) {
```
vguzmanp: Why is the ViewModel binded inconditionally but the view not?

VAS.Core/MVVMC/BindingContext.cs
```diff
@@ -0,0 +1,68 @@
...
+		public BindingContext ()
+		{
+			Bindings = new List<Binding> ();
+		}
+
+		protected override void DisposeManagedResources ()
+		{
+			foreach (Binding binding in Bindings) {
+				binding.Dispose ();
+			}
```
vguzmanp: Clear the list too

VAS.Core/ViewModel/TimelineVM.cs
```diff
@@ -269,11 +296,35 @@ void AddTimelineEventVM (TimelineEventVM viewModel)
...
 
 		void RemoveTimelineEventVM (TimelineEventVM viewModel)
 		{
+			foreach (Player player in viewModel.Model.Players) {
+				playerToTimeline [player].ViewModels.Remove (viewModel);
+			}
 			eventTypeToTimeline [viewModel.Model.EventType.Name].ViewModels.Remove (viewModel);
 		}
+
+		void FullTimeline_PropertyChanged (object sender, PropertyChangedEventArgs e)
```
ylatuya: Rename to HandleFullTimelinePropertyChanged

VAS.Core/MVVMC/PropertyBinding.cs
```diff
@@ -0,0 +1,122 @@
...
+		{
+			ViewModel.PropertyChanged += HandlePropertyChanged;
+		}
+
+		protected void WritePropertyValue (T val)
+		{
+			propertySet (ViewModel, val);
+		}
+
+		abstract protected void WriteViewValue (T val);
```
vguzmanp: Did we have a coding style rule about order here? I think it's better `protected abstract void`

VAS.Tests/MVVMC/TestBinding.cs
```diff
@@ -0,0 +1,79 @@
...
+		public void TestUnbindViewIndisposed ()
+		{
+			var viewModel = new ViewModelBase<BindableBase> ();
+
+			DummyBinding binding = new DummyBinding ();
+			binding.ViewModel = viewModel;
+			binding.Dispose ();
+
+			Assert.IsTrue (binding.viewUnbinded);
+		}
```
vguzmanp: Add a test for the case when you set a new viewmodel when already binded. It should rebind to the new VM, right?

VAS.UI.Helpers.Gtk2/Bindings/BinginsExtensions.cs
```diff
@@ -0,0 +1,130 @@
...
+using VAS.Core.MVVMC;
+using Image = VAS.Core.Common.Image;
+
+namespace VAS.UI.Helpers.Bindings
+{
+	public static class BinginsExtensions
+	{
+
+		/// <summary>
+		/// Gets the binding context for a given widget
```
vguzmanp: Add the comment that it creates it if it doesn't exists

VAS.UI.Helpers.Gtk2/Bindings/ButtonBinding.cs
```diff
@@ -0,0 +1,79 @@
...
+			button.Clicked -= HandleClicked;
+		}
+
+		void BindSensitiveness ()
+		{
+			button.Sensitive = Command.CanExecute ();
+			sensitiveHandler = (sender, e) => {
+				button.Sensitive = Command.CanExecute ();
+			};
+			Command.CanExecuteChanged += sensitiveHandler;
```
vguzmanp: This is basically the same as the `ButtonHelper.Bind` can't you call it?

VAS.UI.Helpers.Gtk2/ButtonHelper.cs
```diff
@@ -138,41 +130,39 @@ static public void SetImage (this Button button, Pixbuf icon)
...
 		}
 
-		/// <summary>
-		/// Bind a button to  the specified button, command and parameter.
-		/// </summary>
-		/// <param name="button">Button.</param>
-		/// <param name="command">Command.</param>
-		/// <param name="parameter">Parameter.</param>
-		public static void Bind (this Button button, Command command, object parameter = null)
+		public static void Bind (this Button button, Command command)
```
vguzmanp: You removed the comment

VAS.UI.Helpers.Gtk2/ButtonHelper.cs
```diff
@@ -138,41 +130,39 @@ static public void SetImage (this Button button, Pixbuf icon)
...
-				command.CanExecuteChanged -= handler;
-			};
-			button.Clicked += (sender, e) => {
-				var radioButton = sender as RadioButton;
-				if (radioButton != null && radioButton.Active == false) {
-					return;
-				}
-				command.Execute (parameter);
-			};
+		public static void LinkToggleButtons (params ToggleButton [] buttons)
```
vguzmanp: Missing comment

VAS.UI.Helpers.Gtk2/ButtonHelper.cs
```diff
@@ -138,41 +130,39 @@ static public void SetImage (this Button button, Pixbuf icon)
...
+		{
+			for (int i = 0; i < buttons.Length; i++) {
```
vguzmanp: Why not use a foreach?

VAS.UI.Helpers.Gtk2/ButtonHelper.cs
```diff
@@ -138,41 +130,39 @@ static public void SetImage (this Button button, Pixbuf icon)
...
-		/// <param name="command">Command.</param>
-		/// <param name="parameter">Parameter.</param>
-		public static void Bind (this Button button, Command command, object parameter = null)
+		public static void Bind (this Button button, Command command)
 		{
-			button.SetImage (command.Icon?.Value);
-			button.Label = command.Text;
-			button.TooltipText = command.ToolTipText;
+			button.Configure (command.Icon?.Value, command.Text, command.ToolTipText, (sender, e) => command.Execute ());
+		}
```
vguzmanp: I find this a bit counter intuitive. Even if we have the bindings, with binding contexts, etc; the `Bind` method should still bind the command, so that the changes in canExecute change the `Sensitive` property.

I think that, either we remove the `Bind` method from here, and go full `ButtonBinding`, or we keep the old behaviour for this method.

VAS.UI.Helpers.Gtk2/ButtonHelper.cs
```diff
@@ -138,41 +130,39 @@ static public void SetImage (this Button button, Pixbuf icon)
...
-			};
+		public static void LinkToggleButtons (params ToggleButton [] buttons)
+		{
+			for (int i = 0; i < buttons.Length; i++) {
+				ToggleButton button = buttons [i];
+				button.Toggled += (sender, e) => {
+					if (button.Active) {
+						foreach (var otherButton in buttons) {
+							if (otherButton != button) {
+								otherButton.Active = false;
```
vguzmanp: Isn't this what RadioButtons do automatically? I believe most our "toggle" behaviour is implemented via RadioButtons for this reason

VAS.UI.Gtk2/UI/Component/LimitationWidget.cs
```diff
@@ -80,11 +83,7 @@ public void SetViewModel (object viewModel)
...
 		void HandlePropertyChangedEventHandler (object sender, PropertyChangedEventArgs e)
 		{
-			if (e.PropertyName == nameof (ViewModel.Count)) {
-				count_label.Text = ViewModel.Count.ToString ();
-				CheckCount ();
-			} else if (e.PropertyName == nameof (ViewModel.Maximum)) {
-				limit_label.Text = ViewModel.Maximum.ToString ();
+			if (ViewModel.NeedsSync (e, nameof (ViewModel.Count)) || ViewModel.NeedsSync (e, nameof (ViewModel.Maximum))) {
 				CheckCount ();
 			} else if (e.PropertyName == nameof (ViewModel.Enabled)) {
```
vguzmanp: else if will not get executed on Sync

VAS.Core/MVVMC/ViewModelBase.cs
```diff
@@ -67,6 +78,14 @@ public bool NeedsSync (string propertyNameChanged, string propertyNameToCheck, o
 			return false;
 		}
 
+		/// <summary>
+		/// Force a sync of all properties as when the ViewModel is set for the first time in a View.
+		/// </summary>
+		public void Sync ()
```
vguzmanp: Shouldn't this go to `BindableBase`?

VAS.Core/Events/GenericEvents.cs
```diff
@@ -118,5 +126,17 @@ public class SaveEvent<T> : ReturningValueEvent
 		/// <value>The object.</value>
 		public T Object { get; set; }
 	}
+
+	/// <summary>
+	/// Event sent to request saving a <typeparam name="T">.
+	/// </summary>
+	public class ReplaceEvent<T> : ReturningValueEvent
```
vguzmanp: I don't understand these two events. `ReplaceEvent` is the same as Update, if the `Object` ID is the same as an existing one, isn't it?

And I don't understand the usage of `ChangeEvent`

VAS.Core/ViewModel/DashboardVM.cs
```diff
@@ -29,6 +30,129 @@ namespace VAS.Core.ViewModel
...
+		public Command<bool> TogglePopupWindow {
+			get;
+			private set;
+		}
+
+		/// <summary>
+		/// Gets the command to change the dashboard fit mode.
+		/// </summary>
+		/// <value>The change fit mode.</value>
+		public Command<FitMode> ChangeFitMode {
```
vguzmanp: I think we should split this in a DashboardEditorVM, containing a DashboardVM. The main usage of a dashboard isn't the edition, so I think these commands shouln't be here.

With an editor VM, these commands can be in the editor, and modify the needed properties of the DashboardVM.

VAS.Core/ViewModel/DashboardVM.cs
```diff
@@ -129,6 +252,45 @@ public class DashboardVM : TemplateViewModel<Dashboard, DashboardButton, Dashboa
...
+			get {
+				return string.Join ("-", Model.GamePeriods);
+			}
+			set {
+				try {
+					Model.GamePeriods = new ObservableCollection<string> (value.Split ('-'));
+				} catch {
+					App.Current.Dialogs.ErrorMessage (Catalog.GetString ("Could not parse game periods."));
+				}
+			}
```
vguzmanp: Why not expose the GamePeriods directly, and let the view show them as it wants?

VAS.Services/Controller/DashboardEditorController.cs
```diff
@@ -0,0 +1,142 @@
...
+			App.Current.EventsBroker.UnsubscribeAsync<ReplaceDashboardFieldEvent> (HandleReplaceField);
+			base.Stop ();
+		}
+
+		public override void SetViewModel (IViewModel viewModel)
+		{
+			dashboardVM = (DashboardVM)(viewModel as dynamic);
+		}
+
+		void ReplaceField (FieldPositionType fieldPosition, Image background = null)
```
vguzmanp: wrong order

VAS.Services/Controller/DashboardEditorController.cs
```diff
@@ -0,0 +1,142 @@
...
+			Image background = await App.Current.Dialogs.OpenImage ();
+			if (background == null) {
+				return;
+			}
+
+			background.ScaleInplace (Constants.MAX_BACKGROUND_WIDTH, Constants.MAX_BACKGROUND_HEIGHT);
+			ReplaceField (arg.Field, background);
+		}
+
+		protected virtual Task HandleResetField (ResetDashboardFieldEvent arg)
```
vguzmanp: Why return a Task if it's not async?

VAS.Services/Controller/DashboardEditorController.cs
```diff
@@ -0,0 +1,142 @@
...
+		protected virtual async Task HandleRemoveLink (DeleteEvent<ActionLink> evt, bool force = false)
+		{
+			ActionLink link = evt.Object;
+			if (link == null) {
+				return;
+			}
+
+			string msg = string.Format ("{0} {1} ?",
+										Catalog.GetString ("Do you want to delete: "), link);
+			if (force || await App.Current.Dialogs.QuestionMessage (msg, null, this)) {
```
vguzmanp: Other methods (like `HandleDeleteButton`) don't have a `force` parameter.

VAS.Services/Controller/DashboardEditorController.cs
```diff
@@ -0,0 +1,142 @@
...
+				link.SourceButton.ActionLinks.Remove (link);
+			}
+		}
+
+		protected virtual DashboardButton CreateButton (string buttonType)
```
vguzmanp: As `HandleCreateButton` and `CreateButton` are both protected, it would be clearer if they were together in the code.

VAS.Services/ViewModel/TemplateManagerViewModel.cs
```diff
@@ -42,64 +42,86 @@ public TemplatesManagerViewModel ()
...
 			get;
-			set;
+			protected set;
 		}
 
 		/// <summary>
 		/// Command to export the currently loaded template.
 		/// </summary>
-		public Task<bool> Export ()
+		protected virtual Task<bool> Export ()
```
vguzmanp: Shouldn't it be async?

VAS.Core/ViewModel/DashboardVM.cs
```diff
@@ -197,8 +199,15 @@ public DashboardVM ()
...
 		public Time CurrentTime {
-			get;
-			set;
+			get {
+				return currentTime;
+			}
+			set {
+				currentTime = value;
+				foreach (var timedVM in ViewModels.OfType<TimedDashboardButtonVM> ()) {
+					timedVM.CurrentTime = value;
```
vguzmanp: The TaggingController was updating also `TimerButtonVM`. Why this one isn't?

VAS.Tests/Core/ViewModel/TestDashboardVM.cs
```diff
@@ -76,7 +77,17 @@ public void TestViewInstanceCreation ()
 
 			Assert.DoesNotThrow (() => analysisButtonView = (AnalysisEventButtonView)App.Current.ViewLocator.Retrieve (dashboard.ViewModels [0].View));
 			Assert.DoesNotThrow (() => timerButtonView = (TimerButtonView)App.Current.ViewLocator.Retrieve (dashboard.ViewModels [5].View));
+		}
+
+		[Test]
+		public void TestProgateCurrentTime ()
```
vguzmanp: typo `Progate` -> `Propagate`

VAS.Drawing/Widgets/DashboardCanvas.cs
```diff
@@ -481,36 +500,16 @@ void HandlePropertyChanged (object sender, PropertyChangedEventArgs e)
...
-			foreach (ActionLinkView ao in Objects.OfType<ActionLinkView> ()) {
-				ao.Visible = ViewModel.ShowLinks;
-				ao.ResetDrawArea ();
+			if (ViewModel.NeedsSync (e, nameof (ViewModel.FitMode))) {
+				HandleSizeChangedEvent ();
 			}
-			ClearSelection ();
-			widget?.ReDraw ();
 		}
+
```
vguzmanp: extra whitespace

VAS.Core/ViewModel/PlayerTimelineVM.cs
ylatuya: fixed

VAS.Core/ViewModel/PlayerTimelineVM.cs
ylatuya: It's not redundant, some of the generic classes requires a parameterless constructor

VAS.Core/ViewModel/PlayerTimelineVM.cs
ylatuya: Moved to NestedViewModel

VAS.Core/ViewModel/TimelineVM.cs
ylatuya: Fixed

VAS.Core/ViewModel/TimelineVM.cs
ylatuya: Fixed

VAS.Tests/Core/ViewModel/TestTimelineVM.cs
ylatuya: Not sure what you mean

VAS.Core/MVVMC/BindableBase.cs
```diff
@@ -179,7 +191,7 @@ protected virtual void CollectionChanged (object sender, NotifyCollectionChanged
 					element.PropertyChanged += ForwardPropertyChanged;
 				}
 			}
-			RaisePropertyChanged ("Collection", this);
+			RaisePropertyChanged ($"Collection_{collectionToPropertyName [sender as INotifyCollectionChanged]}", this);
```
ylatuya: Pending

VAS.Drawing/CanvasObjects/Dashboard/AnalysisEventButtonView.cs
ylatuya: fixed

VAS.Core/MVVMC/Binding.cs
ylatuya: There one binding for one view, they are linked together, while a binding could use different ViewModel's. The view should be binded in the constructor, only once, but since BindView() is a virtual we need to delay the bind until the ViewModel is set for the first time.

VAS.Core/MVVMC/BindingContext.cs
ylatuya: fixed

VAS.Core/MVVMC/PropertyBinding.cs
ylatuya: fixed

VAS.Tests/Core/ViewModel/TestTimelineVM.cs
vguzmanp: Nevermind, I see it now.

VAS.Tests/MVVMC/TestBinding.cs
ylatuya: fixed

VAS.UI.Helpers.Gtk2/ButtonHelper.cs
ylatuya: Radio buttons can't use images, only toggle buttons, an example using toogle buttons as groups are the dashboards button configuring the DashboardMode

VAS.Core/MVVMC/Binding.cs
vguzmanp: I feel that several ViewModels would be a problem.

For example, using a `TextView`, you bind it to one VM, and then to a second one.
The view will react to both, but will only write to the last one (because it will be the one in the `ViewModel` property here).

VAS.UI.Helpers.Gtk2/ButtonHelper.cs
vguzmanp: Our "Tab" buttons (like the ones to choose Events/Presentations/Filters) are radioButtons internally, and they use images.

VAS.Core/MVVMC/Binding.cs
ylatuya: It will never be binded to 2 VM's at the same time, it's just the ViewModel can change but the View can't

VAS.UI.Helpers.Gtk2/ButtonHelper.cs
ylatuya: fixed

VAS.UI.Helpers.Gtk2/ButtonHelper.cs
ylatuya: fixed

VAS.UI.Helpers.Gtk2/ButtonHelper.cs
ylatuya: But they don't have the Toggled state, like toggle buttons do, so it's just an image, that you can change but you can't make it "toggled"

VAS.UI.Gtk2/UI/Component/LimitationWidget.cs
ylatuya: fixed

VAS.Core/MVVMC/Binding.cs
vguzmanp: I don't follow you here... According to this test (https://github.com/fluendo/VAS/pull/224/files/6dfe633d33c1ac38bed0b3b0abdbe1a61803f65f..17f6e3011cab064edcf61a14cc82cabc08c3f3a2#diff-bc9d1cb90aac76c6018f906c63f73e8cR86) it's binded to 2 VM at the same time, as we don't unbind from the first.

VAS.Tests/Core/ViewModel/TestDashboardVM.cs
ylatuya: fixed

VAS.Core/ViewModel/DashboardVM.cs
ylatuya: And it was doint it twice since a TimerButtonVM is a TimedDashboardButtonVM too

VAS.Services/ViewModel/TemplateManagerViewModel.cs
ylatuya: Why? There are not awaits in the code, we return tasks directly

VAS.Services/Controller/DashboardEditorController.cs
ylatuya: fixed

VAS.Services/Controller/DashboardEditorController.cs
ylatuya: fixed

VAS.Services/Controller/DashboardEditorController.cs
ylatuya: fixed

VAS.Core/ViewModel/DashboardVM.cs
ylatuya: But those properties dictates how the dashboard is drawn, they belong  to the Dashboard, wether it's inside an editor or not.

vguzmanp: The properties do, but I think the commands don't.

VAS.Core/Events/GenericEvents.cs
ylatuya: Those events are not used and chould be rmeove, I left them by mistake from the firsts iterations

VAS.Core/MVVMC/Binding.cs
ylatuya: It's only binded to one, BindViewModel must take care of disconnecting the old one and connecting the new one. We can split it in 2 phases to make it more clear

VAS.UI.Helpers.Gtk2/ButtonHelper.cs
vguzmanp: You have the Toggled event, just like here, and the Active flag, just like here.

from [the radio button docs](https://developer.gnome.org/gtk2/stable/GtkRadioButton.html)
> When an unselected button in the group is clicked the clicked button receives the "toggled" signal, as does the previously selected button. Inside the "toggled" handler, gtk_toggle_button_get_active() can be used to determine if the button has been selected or deselected. 

Why can't you use the `Active` as `Toggled`?

VAS.Core/MVVMC/ViewModelBase.cs
vguzmanp: Why not? the typical pattern here is:

```
set {
	if (field != null) {
		field.PropertyChanged -= HandlePropertyChanged;
	}
	field = value;
	if (field != null) {
		field.PropertyChanged += HandlePropertyChanged;
		field.Sync ();
	}
}
```

There is nothing ViewModel specific there, only PropertyChanged, which comes from BindableBase

VAS.Core/MVVMC/Binding.cs
vguzmanp: Yes, if we force children to implement `UnbindViewModel`, we don't risk having some class leaving the old viewmodel connected.

VAS.Services/ViewModel/TemplateManagerViewModel.cs
vguzmanp: 👌

VAS.Drawing/CanvasObjects/Dashboard/ActionLinkView.cs
```diff
@@ -138,8 +137,8 @@ public bool CanLink (LinkAnchorView dest)
 			}
 
 			/* Check if this link will result into a duplicated link */
-			foreach (ActionLink link in Source.Button.Button.ActionLinks) {
-				if (link.DestinationButton == dest.Button.Button &&
+			foreach (ActionLink link in Source.Button.ButtonVM.Model.ActionLinks) {
```
ylatuya: fixed

VAS.Drawing/Widgets/DashboardCanvas.cs
ylatuya: fixed

VAS.UI.Helpers.Gtk2/ButtonHelper.cs
ylatuya: fixed

VAS.UI.Helpers.Gtk2/Bindings/BinginsExtensions.cs
ylatuya: fixed

VAS.Core/ViewModel/DashboardVM.cs
ylatuya: In the analysis window you can still change the display mode, and you are not inside the editor, you will need to have a ProjectVM, a DashboardEditorVM from the ProjectVM.Dashboard and the DashboardVM

vguzmanp: Ok, the `ChangeFitMode` command is about the dashboard, but the `AddButton` etc aren't. My comment was in general about the commands, not only the changeFitMode.

VAS.Drawing/Widgets/DashboardCanvas.cs
```diff
@@ -454,6 +453,26 @@ void RemoveButton (DashboardButtonVM vm)
 			buttonsDict.Remove (vm.Model);
 		}
 
+		void UpdateMode ()
+		{
+			ObjectsCanMove = ViewModel.Mode == DashboardMode.Edit;
+			ClearSelection ();
```
ylatuya: That's a problem we still have in our custom widgets, the selections int the View are diferent from the selections in the VM's. The Controller has no knoledge of that so that's why this needs to be done here too until it's correctly done everywhere.

vguzmanp: 👌 
The correct way would be subscribing to the selectionChanged event and modifying the view in the selection accordingly, right?

VAS.Core/MVVMC/ViewModelBase.cs
ylatuya: Even though we discussed that in the chat, if we move Sync () we also need to move NeedSync. A bindable per-se doesn't require to sync itself. I will leave it like that and we can change it the first time a BindableBase that's not a ViewModelBase needs it :D

PR #225, opened 2017-03-02
RA-213: modify the event duration on the timeline

aameijide:


VAS.Drawing/CanvasObjects/Timeline/TimelineEventView.cs
```diff
@@ -89,6 +92,17 @@ protected virtual void DrawBorders (IDrawingToolkit tk, double start, double sto
...
+				if (cursor != CursorType.DoubleArrow) {
+					App.Current.EventsBroker.Publish (new SetCursorEvent { CursorType = CursorType.DoubleArrow });
+					cursor = CursorType.DoubleArrow;
+				}
+			} else {
+				if (cursor != CursorType.Arrow) {
+					App.Current.EventsBroker.Publish (new SetCursorEvent { CursorType = CursorType.Arrow });
+					cursor = CursorType.Arrow;
+				}
+			}
```
vmartos: Why this logic to change the CursorType, is in the DrawBorders method?

VAS.Drawing/Widgets/PlaysTimeline.cs
```diff
@@ -439,5 +447,10 @@ void HandleChildsVMCollectionChanged (object sender, NotifyCollectionChangedEven
 				}
 			}
 		}
+
+		void HandleSetCursorEvent (SetCursorEvent obj)
+		{
+			widget.SetCursor (obj.CursorType);
```
vmartos: Also i don't like the idea of sending events to react on Cursor changes. This PlaysTimeline knows the type and if it's childnodes have the dragginMode enabled or no, so you can override on CursorMoved and depending on the selection and dragMode you just can set the type of Cursor you wish to show.

VAS.Drawing/Widgets/PlaysTimeline.cs
```diff
@@ -313,7 +321,7 @@ protected override void StartMove (Selection sel)
 
 		protected override void StopMove (bool moved)
 		{
-			widget.SetCursor (CursorType.Arrow);
+			widget.SetCursor (CursorType.DoubleArrow);
```
vmartos: this set the cursors to DoubleArrow by clicking in the timeline outside a TimelineEvent

VAS.Drawing/Widgets/PlaysTimeline.cs
```diff
@@ -178,6 +172,22 @@ protected void AddTimeline (TimelineView timelineView, IViewModel viewModel)
 			}
 		}
 
+		protected override void CursorMoved (Point coords)
+		{
+			base.CursorMoved (coords);
+			if (HighlightedObject is TimelineEventView) {
```
vmartos: Timers do not change to have a double arrow. Maybe here you can cast to TimeNodeView and get the DraggingMode, we don't want to use a DoubleArrow cursor for elements that we can't drag. Also take into account that in matchdata line in the timeline the elements can be dragged. you should set the NodeDraggingMode to none to the elements of type RATimelineEventWithIconView.

Also if we can at this point differentiate between different dragging modes All, Borders, Segment could be cool! but if it's a lot of work we can just change the cursor to DraggingMode.All TimeNode Types that in riftanalyst will be almost all.

vmartos:
👍

PR #226, opened 2017-03-02
Fix Windows Ping Exception: Check also using NetWorkTime

vmartos:


VAS.Core/Common/Utils.cs
```diff
@@ -218,11 +220,48 @@ public static bool CheckNetworkConnection ()
 				PingReply reply = myPing.Send (host, timeout, buffer, pingOptions);
 				return (reply.Status == IPStatus.Success);
 			} catch (Exception) {
-				return false;
+				try {
+					//Try with NTP server
+					GetNetworkTime ();
+					return true;
```
vguzmanp: Compare that networkTime is > new DateTime (1900, 1, 1, 0, 0, 0, DateTimeKind.Utc)

PR #228, opened 2017-03-14
RA-574: The team name doesn't change

aameijide:


VAS.Services/Controller/TemplatesController.cs
```diff
@@ -369,7 +369,7 @@ protected bool IsChildSelection (IEnumerable<TViewModel> selectedViewModels)
 		{
 			TModel template = evt.Object;
 			bool force = evt.Force;
-			TViewModel templateViewmodel = ViewModel.ViewModels.FirstOrDefault (vm => vm.Model.Equals (template));
+			TViewModel templateViewmodel = ViewModel.ViewModels.FirstOrDefault (vm => vm.Model.ID.Equals (template.ID));
```
vmartos: I don't get why it should compare by ID if the base of Team : StorableBase compares already with ID

VAS.Core/Store/Templates/Team.cs
```diff
@@ -108,5 +108,23 @@ public virtual ITemplate Copy (string newName)
...
+			if (!(obj is Team)) {
+				return false;
+			}
+
+			Team other = obj as Team;
+
+			if (!base.Equals (obj)) {
+				return false;
+			}
+			if (this.Name != other.Name) {
```
vmartos: What was the original issue! why this extra check by name is added?

aameijide: In templatesController (line 393) this line was not working:
`templateViewmodel.Model = template;`
And that was because - before assign the value - it checks if there are equals and if true, it doesn't change the value. As equals only was checking ID and template (with a name or region changed) and templateVM.Model had the same ID value, templateVM.Model was never updated.
For that reason Name and region (in RA) are also checked, as are the ones that we can change.

VAS.Services/Controller/TemplatesController.cs
```diff
@@ -389,7 +389,7 @@ protected bool IsChildSelection (IEnumerable<TViewModel> selectedViewModels)
 							// When is a new template, we should get the VM again, because previously was null
 							templateViewmodel = ViewModel.ViewModels.FirstOrDefault (vm => vm.Model.Equals (template));
 						} else {
-							// Update the ViewModel with the model clone used for editting. If it was a new one isn't necessary
+							// Update the ViewModel with the model clone used for editing. If it was a new one isn't necessary
 							templateViewmodel.Model = template;
```
aameijide: That is the line I was referring in the comment in Team.cs file

VAS.Services/Controller/TemplatesController.cs
aameijide: Same case that Team.cs. Now equals also compares that name and region (in RA) are equal. As template can have one or both properties changed, `vm.Model.Equals(template)` will always false

vmartos:
just rembember the shield/image problema before merging

PR #240, opened 2017-03-30
Ra 681

vmartos:


VAS.Core/Interfaces/GUI/IBusyDialog.cs
```diff
@@ -45,6 +46,15 @@ public interface IBusyDialog
 		/// <param name="action">The action to run in the background.</param>
 		/// <param name="pulseIntervalMS">The pulse interval in milliseconds.</param>
 		void ShowSync (Action action, uint pulseIntervalMS = 0);
+
+		/// <summary>
+		/// Show the dialog synchronously and run the <paramref name="asyncAction"/> in background.
```
vguzmanp: Do you mean `asynchronously`? If the Dialog is synchronous, the method shouldn't be called like that

VAS.Services/Controller/TemplatesController.cs
```diff
@@ -172,13 +172,18 @@ public override IEnumerable<KeyAction> GetDefaultKeyActions ()
 		/// Removes the selected children.
 		/// </summary>
 		/// <param name="vm">Vm.</param>
-		protected virtual async Task RemoveSelectedChildren (TViewModel vm)
+		protected virtual void RemoveChildsFromTemplate (TViewModel template, IEnumerable<TChildViewModel> childs)
```
vguzmanp: Don't we want to make this async too?

VAS.UI.Gtk2/UI/Dialog/BusyDialog.cs
```diff
@@ -103,6 +103,36 @@ public void ShowSync (System.Action action, uint pulseIntervalMS = 100)
...
+						}
+					});
+				}
+			});
+			Monitor.Enter (lockObject);
+			task.Start ();
+			timeout = GLib.Timeout.Add (pulseIntervalMS, OnTimeout);
+			Run ();
+			if (ex != null)
+				throw ex;
```
vguzmanp: Add braces

VAS.UI.Gtk2/UI/Dialog/BusyDialog.cs
```diff
@@ -103,6 +103,36 @@ public void ShowSync (System.Action action, uint pulseIntervalMS = 100)
...
+			Exception ex = null;
+			object lockObject = this.lockObject = new object ();
+
+			if (pulseIntervalMS == 0) {
+				pulseIntervalMS = 100;
+			}
+			Task task = new Task (async () => {
+
+				try {
+					await action ();
```
vguzmanp: What do we win with this? It's kind of an override with different parameters, but it does exactly the same as `ShowSync`, no?

VAS.Services/Controller/TemplatesController.cs
vmartos: It is not necessary because now doesn't have any asynchronous operation in this method and overrided method in Rift. Due to the use of a BusyDialog when removing childs.

VAS.Core/Interfaces/GUI/IBusyDialog.cs
vmartos: Yes maybe the name of the method shouldn't be ShowAsync, because as it is a Dialog should be synchronous. The idea behind this new method is to be able to use BusyDialog in asynchronoys methods. Maybe I can rename ShowAsync to ShowSync but recieving a Func<Task> instead of Action.

vguzmanp: Yes, I think it's better. Maybe you will have some problems with the two, as they are castable between them or something like that. We have the same in Command I think

PR #241, opened 2017-04-03
RA-760

dfernandezramos:


VAS.Core/Common/Image.cs
```diff
@@ -105,9 +109,26 @@ protected override Pixbuf Scale (Pixbuf pix, int maxWidth, int maxHeight)
...
+		/// </summary>
+		/// <param name="filename">Filename.</param>
 		public override void Save (string filename)
 		{
-			Value.Save (filename, "png");
+			//HACK: Force gdk_pixbuf_save_utf8 call if windows OS. Otherwhise call gdk_pixbuf_save
+			if (Environment.OSVersion.Platform == PlatformID.Win32NT) {
+				IntPtr error = IntPtr.Zero;
+				IntPtr native_filename = GLib.Marshaller.StringToPtrGStrdup (filename);
+				IntPtr native_type = GLib.Marshaller.StringToPtrGStrdup ("png");
```
aameijide: Move "png" to a constant in order to use it here and in line 130

VAS.Core/Common/Image.cs
```diff
@@ -105,9 +109,26 @@ protected override Pixbuf Scale (Pixbuf pix, int maxWidth, int maxHeight)
...
 
+		/// <summary>
+		/// Save the image in the specified path.
+		/// </summary>
+		/// <param name="filename">Filename.</param>
 		public override void Save (string filename)
 		{
-			Value.Save (filename, "png");
+			//HACK: Force gdk_pixbuf_save_utf8 call if windows OS. Otherwhise call gdk_pixbuf_save
+			if (Environment.OSVersion.Platform == PlatformID.Win32NT) {
```
aameijide: Use better #if OSTYPE_WINDOWS (also on the using and the Dll imported method). At that way, this code will only compile in windows and not for others OS.

aameijide: As @ylatuya commented, it's better to use `if (Environment.OSVersion.Platform == PlatformID.Win32NT)` instead of `#if OSTYPE_WINDOWS` in order to have a portable library.

VAS.Core/GdkGlue.cs
```diff
@@ -0,0 +1,46 @@
...
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using System;
+using Gdk;
+using System.Runtime.InteropServices;
```
aameijide: order usings

PR #242, opened 2017-04-05
Lon 961

SandraPG85:
** ScreenState extended to add a Freeze state that modal views call to instead of Hide.
** StateController extended to wait for a modal transition to be completed
** EventController merged with CoreEventsController
** Fix popup windows in dashboard events and substitution events from tree view


VAS.Core/Common/EventEditionSettings.cs
```diff
@@ -0,0 +1,49 @@
...
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+namespace VAS.Core.Common
+{
+	/// <summary>
+	/// Settings of the play events edition
```
vguzmanp: Can you give a bit more info about the class? Intended usage, etc

VAS.Core/Common/StateController.cs
```diff
@@ -372,7 +383,7 @@ public bool UnRegister (string transition)
 				if (!await navigationStateStack [i].ScreenState.HideState ()) {
 					return false;
 				}
-				if (!await navigationStateStack [i].ScreenState.UnloadState ()) {
+				if (!await navigationStateStack [i].Unload ()) {
```
vguzmanp: Why is this one different from the others? Shouldn't it be in ScreenState?

VAS.Services/Controller/CoreEventsController.cs
```diff
@@ -38,59 +35,48 @@
 
 namespace VAS.Services.Controller
 {
-	// FIXME: Merge with the EventsController
```
vguzmanp: Is this actually fixed?

VAS.UI.Gtk2/UI/TreeView/TimelineEventsTreeView.cs
```diff
@@ -139,6 +139,9 @@ protected override void HandleTreeviewSelectionChanged (object sender, EventArgs
 			if (Selection.CountSelectedRows () == 0) {
 				ViewModel.UnloadEvents ();
 			}
+
+			// update selection
+			ViewModel.FullTimeline.Selection.Replace (events);
```
vguzmanp: How is this related to the rest?

VAS.Services/Controller/EventsController.cs
```diff
@@ -138,6 +293,101 @@ void HandleFiltersChanged ()
...
+		}
+
+		void HandlePlaylistElementLoaded (PlaylistElementLoadedEvent e)
+		{
+			if (e.Element is PlaylistPlayElement) {
+				LoadedPlay = (e.Element as PlaylistPlayElement).Play;
+			} else {
+				LoadedPlay = null;
+			}
+		}
```
vguzmanp: From all these changes, what are movements from the other controller and what are actual changes?

VAS.Core/Interfaces/IEventEditorService.cs
```diff
@@ -0,0 +1,39 @@
...
+using VAS.Core.Common;
+using VAS.Core.Store;
+using VAS.Core.ViewModel;
+
+namespace VAS.Core.Interfaces
+{
+	/// <summary>
+	/// Service that offers graphical event edition
+	/// </summary>
+	public interface IEventEditorService
```
vguzmanp: We have a `Service` interface, that works differently to this one. In general, services are application-wide controllers, but this one should be a normal controller, shouldn't it?

VAS.Core/Common/StateController.cs
SandraPG85: Notify about the completion of the navigation is responsibility of the navigation state instead of the screen state because is a more oriented in the navigation.

VAS.Services/Controller/CoreEventsController.cs
SandraPG85: I checked with Andoni the events that can be moved to the events controller, the remaining step should be find a better name for the CoreEventsController

VAS.UI.Gtk2/UI/TreeView/TimelineEventsTreeView.cs
SandraPG85: this fixes a bug in the selection, since it was all time empty.

VAS.Services/Controller/EventsController.cs
SandraPG85: adding the handleEditevent function

VAS.Core/ViewModel/TimelineVM.cs
```diff
@@ -54,6 +55,7 @@ public TimelineVM ()
 			FullTimeline = CreateFullTimeline ();
 			FullTimeline.ViewModels.CollectionChanged += HandleTimelineCollectionChanged;
 			FullTimeline.PropertyChanged += FullTimeline_PropertyChanged;
+			EditionCommand = new Command<TimelineEvent> (HandleEditPlay, (arg) => true);
```
SandraPG85: there is a task to fix the tree view behavior, right know the command is only being used by the execute function, so this can be checked there, right now my answer will be that yes :)

VAS.Core/Common/StateController.cs
vguzmanp: Do the same in all other transitions.

VAS.Core/Common/StateController.cs
```diff
@@ -538,7 +538,9 @@ public override bool Equals (object obj)
 		public async Task<bool> Unload ()
 		{
 			bool result = await ScreenState.UnloadState ();
-			Completion.SetResult (result);
+			if (Completion.Task.Status != TaskStatus.RanToCompletion) {
```
vguzmanp: In which case is the task already completed?

vguzmanp: After our chat: Add a `TechnicalDebt` task, and a FIXME for this, as it's hiding an exception

PR #244, opened 2017-04-07
LON-981: Editing a player does not changes the state of the save button

aameijide:
* Process old and new items when collection changed

### DIFF
--- a/VAS.Tests/MVVMC/TestBindableBase.cs
+++ b/VAS.Tests/MVVMC/TestBindableBase.cs
@@ -16,10 +16,12 @@
 //  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 //
 using System;
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using NUnit.Framework;
 using VAS.Core.Common;
 using VAS.Core.Store;
+using DummyPlayer = VAS.Tests.Utils.PlayerDummy;
 
 namespace VAS.Tests.MVVMC
 {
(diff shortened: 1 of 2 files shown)


VAS.Tests/Core/MVVMC/TestBindableBase.cs
```diff
@@ -0,0 +1,60 @@
...
+//
+using System;
+using NUnit.Framework;
+using VAS.Core.Common;
+using DummyPlayer = VAS.Tests.Utils.PlayerDummy;
+
+namespace VAS.Tests.Core.MVVMC
+{
+	[TestFixture]
+	public class TestBindableBase
```
ylatuya: There is already a test file here https://github.com/fluendo/VAS/blob/7a8adb59fa77a32a4f16f632b056b624c93be737/VAS.Tests/MVVMC/TestBindableBase.cs.

Also be more specific with the name of the test because it testing a very specific use case, which is event forwarding in collection replacements.

VAS.Tests/MVVMC/TestBindableBase.cs
```diff
@@ -161,6 +162,38 @@ public void TestIsChangedSetBeforeRaising ()
...
+
+			team.List.Swap (p1, p2);
+
+			team.IsChanged = false;
+			p1.IsChanged = false;
+			p2.IsChanged = false;
+
+			// Action
+			p2.Nationality = "FR";
+			p1.Nationality = "DE";
```
vmartos: you could also subscribe to PropertyChanged in Team and see that the sender should be the players when changing the Nationality, with this you could test exactly that the forwarder is working. With the IsChanged, you could also test it, because it is changed on the Property changed Handler, but the logic could be changed somehow and the test could pass even if the forwarder is not working.

vmartos:
👍

PR #245, opened 2017-04-07
Lon 982

vmartos:


VAS.Core/MVVMC/ViewModelBase.cs
```diff
@@ -75,7 +75,10 @@ public void Sync ()
 
 		protected override void ForwardPropertyChanged (object sender, PropertyChangedEventArgs e)
 		{
-			base.ForwardPropertyChanged (this, e);
+			if (!(sender is IViewModel)) {
+				sender = this;
+			}
+			base.ForwardPropertyChanged (sender, e);
```
ylatuya: Missing unit test

VAS.Core/MVVMC/NestedSubViewModel.cs
```diff
@@ -71,6 +76,14 @@ public abstract class NestedSubViewModel<TModel, TViewModel, TModelChild, TVMChi
 		protected CollectionViewModel<TModelChild, TVMChild> SubViewModel {
 			get;
 			set;
-		} = new CollectionViewModel<TModelChild, TVMChild> ();
+		}
+
+		protected override void ForwardPropertyChanged (object sender, System.ComponentModel.PropertyChangedEventArgs e)
```
vguzmanp: Add `using System.ComponentModel`

VAS.Core/MVVMC/ViewModelBase.cs
```diff
@@ -93,6 +99,31 @@ public class ViewModelBase<T> : ViewModelBase, IViewModel<T> where T : INotifyPr
...
+
+		protected override void ForwardPropertyChanged (object sender, PropertyChangedEventArgs e)
+		{
+			Log.Verbose ($"ViewModelBase<T>{this} - Reception Forward PropertyName: {e.PropertyName} with sender: {sender}");
+			if (!(sender is IViewModel)) {
+				if (sender == Model) {
+					Log.Verbose ($"ViewModelBase<T> {this} - Changing Forward sender to {this}");
+					sender = this;
+				} else {
+					//Change sender if sender is a Property of the Model
```
vguzmanp: If sender is a property of the Model, shouldn't this be done in the Model?

VAS.Core/Store/ActionLink.cs
```diff
@@ -145,7 +145,7 @@ public override int GetHashCode ()
 
 		public override string ToString ()
 		{
-			return string.Format ("{0} -> {1}", SourceButton.Name, DestinationButton.Name);
+			return string.Format ("{0} -> {1}", SourceButton?.Name, DestinationButton?.Name);
```
vguzmanp: Why would `SourceButton` and `DestinationButton` be null? Are we hiding a real error here?

VAS.Tests/MVVMC/TestViewModelBase.cs
```diff
@@ -48,6 +49,59 @@ public void TestForwardProperty ()
...
+			timer.Nodes.Add (timeNode);
+			viewModel.Model = timer;
+			viewModel.PropertyChanged += (sender, e) => {
+				senderObject = sender;
+			};
+
+			viewModel.ViewModels [0].Name = "test";
+
+			Assert.AreEqual (viewModel.ViewModels [0], senderObject);
+		}
```
vguzmanp: How does this forwarder work with CollectionChanged events? Can you test it?

VAS.Core/MVVMC/NestedSubViewModel.cs
```diff
@@ -71,6 +76,14 @@ public abstract class NestedSubViewModel<TModel, TViewModel, TModelChild, TVMChi
...
+		{
+			if (sender == Model) {
+				sender = this;
+			}
+			base.ForwardPropertyChanged (sender, e);
```
vguzmanp: Isn't this already done in the base method?

vmartos: The problem is that NestedSubViewModel is a NestedViewModel with a model, but NestedViewModel did not inherit from VIewModelBase<Model> so the only possible way to change the sender for NestedSubViewModels is overriding the ViewModelBase forwarder and check is sender equals model

vguzmanp: 👌

VAS.Core/MVVMC/ViewModelBase.cs
vmartos: The BindableBase (Models) did not change the sender besides that sender is null, so this is for special cases where a ViewModel has a Model that has BindableBase Properties inside. Ex:
PlayListElementVM --> Model PlaylistElement --> Property Play (TimelineEvent) --> Name 
If we change the Name inside the Play Property it forwards the propery change with the Play as a sender. then the basic check for sender == Model doesn't work for this special cases.

VAS.Core/Store/ActionLink.cs
vmartos: This was detected when adding verbosing logs on the ForwardPropertyChange in the bindableBase and ViewModelBase. Because a test failed with Null reference because SourceButton was null at this time. Didn't know if SourceButton shouldn't never be null and the test was not arranged correctly

VAS.Core/MVVMC/ViewModelBase.cs
vguzmanp: Hmm in that example, shouldn't we have a `Play` property in the `PlaylistElementVM` that is a `TimelineEventVM` with the same `TimelineEvent` as a Model?.

That way the PropertyChanged would come from the correct VM.

vguzmanp: Or maybe we should create a `ModelBase` to do this sort of things and listen to PropertyChanged of their properties (and change the sender).

VAS.Core/Store/ActionLink.cs
vguzmanp: Can you fix the test instead? I'd prefer this to fail if it's null. I see that in ActionLink it's using them without checks many times

vguzmanp:
Squash and 👍

PR #248, opened 2017-04-12
Lon 968

SandraPG85:


VAS.Services/Controller/TaggingController.cs
```diff
@@ -192,34 +215,33 @@ void HandleVideoPlayerPropertyChanged (object sender, System.ComponentModel.Prop
...
-			}, () => PCardAction (ButtonModifier.Control, player));
-			keyActions.Add (KeyActionControl);
-
-			var KeyActionShift = new KeyAction (new KeyConfig {
-				Name = "Player " + index + " Shift",
-				Key = App.Current.Keyboard.ParseName ("<Shift_L>+" + index.ToString ())
-			}, () => PCardAction (ButtonModifier.Shift, player));
-			keyActions.Add (KeyActionShift);
+			var changedButton = sender as DashboardButtonVM;
+			if (changedButton != null && args.PropertyName == nameof (changedButton.HotKey) &&
```
vmartos: you have NeedSync in DashboardButtonVM so you can do: changedButton.NeedSync(e, nameof (changedButton.HotKey) instead of args.PropertyName == nameof (changedButton.HotKey)

VAS.Services/Controller/TaggingController.cs
```diff
@@ -192,34 +215,33 @@ void HandleVideoPlayerPropertyChanged (object sender, System.ComponentModel.Prop
...
+			
+			KeyTemporalContext tempContext = new KeyTemporalContext { };
+			foreach (var subcategory in buttonVM.Tags) {
+				KeyAction action = new KeyAction (new KeyConfig {
+					Name = subcategory.Value,
+					Key = subcategory.HotKey
+				}, () => HandleSubCategoriesTagging (buttonVM, subcategory));
+				tempContext.AddAction (action);
+			}
+			tempContext.Duration = 500;
```
vmartos: why 500? can we have a Constant in Vas.Constants (TEMP_TAGGING_DURATION) to be able to edit it at one place only? Since it is also used in tagging player in Longomatch Tagging Controller

VAS.Tests/Setup.cs
```diff
@@ -51,6 +51,7 @@ public static void Initialize ()
 			SynchronizationContext.SetSynchronizationContext (new MockSynchronizationContext ());
 			App.Current.DependencyRegistry.Register<IStorageManager, CouchbaseManager> (1);
 			App.Current.DependencyRegistry.Register<IStopwatch, Stopwatch> (1);
+			App.Current.DependencyRegistry.Register<ITimer, VAS.Core.Common.Timer> (1);
```
vmartos: using VAS.Core.Common; if need alias create it

VAS.Core/Common/Registry.cs
```diff
@@ -96,6 +96,40 @@ public void Register (Type interfac, Type type, int priority = 0)
 		}
 
 		/// <summary>
+		/// Register a new element type for a given interface with a priority.
```
vmartos: maybe should be: register a new element with an instance object for a given interface with a priority

VAS.Core/Common/Registry.cs
```diff
@@ -96,6 +96,40 @@ public void Register (Type interfac, Type type, int priority = 0)
...
+		{
+			if (!interfac.IsAssignableFrom(instance.GetType ())) {
+				throw new InvalidOperationException ("Registered instance doe not implement the proper interface");
+			}
+
+			if (!elements.ContainsKey (interfac)) {
+				elements [interfac] = new List<RegistryElement> ();
+			}
+
+			RegistryElement element = new RegistryElement (interfac, priority);
```
vmartos: shouldn't be RegistryElement element = new RegistryElement (instance.GetType(), priority); ?

VAS.Core/Common/Seeker.cs
```diff
@@ -1,10 +1,9 @@
 using System;
 using System.Threading;
-using VAS.Core.Common;
 using VAS.Core.Handlers;
 using VAS.Core.MVVMC;
 using VAS.Core.Store;
-using Timer = System.Threading.Timer;
+using RTimer = System.Threading.Timer;
```
SandraPG85: Because the wrapper Timer added is in the same namespace and if I use the same name looks for the other one.

vmartos: yes but i mean why you set that name, maybe a SystemTimer could be more readable later in code.

PR #250, opened 2017-04-18
Live and fake projects error when capturer not started and click button

SandraPG85:
- AnalysisState should copy the model not the VM since it is disposed
- Currentime is initialised when the dashboardVM is created 
- CapturerTick is checked by the TaggingController to update the CurrentTime of the VM.

### DIFF
--- a/VAS.Services/Controller/CoreEventsController.cs
+++ b/VAS.Services/Controller/CoreEventsController.cs
@@ -83,8 +83,10 @@ namespace VAS.Services.Controller
 		{
 			project = (ProjectVM)(viewModel as dynamic);
 			videoPlayer = (VideoPlayerVM)(viewModel as dynamic);
+
+			// FIXME: Remove the try catch when the new interface is passed instead of IViewModel
 			try {
-				capturer = (ICapturerBin)(viewModel as dynamic);
+				capturer = (ICapturerBin)((viewModel as dynamic).Capturer);
 			} catch {
 			}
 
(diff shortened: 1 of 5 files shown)


VAS.Services/Controller/CoreEventsController.cs
```diff
@@ -84,7 +84,7 @@ public override void SetViewModel (IViewModel viewModel)
 			project = (ProjectVM)(viewModel as dynamic);
 			videoPlayer = (VideoPlayerVM)(viewModel as dynamic);
 			try {
-				capturer = (ICapturerBin)(viewModel as dynamic);
+				capturer = (ICapturerBin)((viewModel as dynamic).Capturer);
 			} catch {
```
aameijide: Do we need this try and catch? Is possible to use: `capturer = (ICapturerBin)((viewModel as dynamic)?.Capturer);`?

SandraPG85: try catch is needed anyway because the capturer is only available when the project is of type fake or live but not video. In the task of tech debt that was planned this will be fixed using an interface.

aameijide: In a video project capturer is null, so it will never enter in the catch, the variable will be setted to null (There's no problem to cast `null` to `ICapturerBin`).

Also, if you still prefer to left it to the tech debt task, please write a FIXME message to not forget it.

PR #253, opened 2017-04-20
Fix navigation from tools menu

SandraPG85:


ylatuya:
Missing unit test for the use case causing this bug.

VAS.Core/Common/StateController.cs
```diff
@@ -106,12 +106,6 @@ public StateController ()
 					}
 				}
 
-				// Precheck: it is possible that hiding a previous state caused a new transition so the 
-				// initial one must be canceled
-				if (initiatedTransition != transition) { 
-					return false;
-				}
```
ylatuya: This was probably added for a reason, isn't it breaking somehow what it was supposed to fix when it was added?

SandraPG85: I added this code during the open/close project bug fixing but at then end it is not necessary because the controller controls the reopen and go home transitions.

PR #255, opened 2017-04-24
Fix more than one tree iter with the same viewmodel

SandraPG85:
It is possible to have two players with the same view model in the events tree view.

### DIFF
--- a/VAS.UI.Gtk2/UI/Common/TreeViewBase.cs
+++ b/VAS.UI.Gtk2/UI/Common/TreeViewBase.cs
@@ -42,7 +42,7 @@ namespace VAS.UI.Common
 		protected const int COL_DATA = 0;
 		protected TreeStore store;
 		TCollectionViewModel viewModel;
-		protected Dictionary<IViewModel, TreeIter> dictionaryStore;
+		protected Dictionary<IViewModel, List<TreeIter>> dictionaryStore;
 		protected Dictionary<INotifyCollectionChanged, TreeIter> dictionaryNestedParent;
 
 		protected TreeModelFilter filter;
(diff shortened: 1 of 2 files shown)


VAS.UI.Gtk2/UI/Common/TreeViewBase.cs
```diff
@@ -525,12 +530,14 @@ protected virtual void HandleViewModelPropertyChanged (object sender, PropertyCh
...
-					if (filter != null) {
-						externalSelected = filter.ConvertChildIterToIter (externalSelected);
-						externalSelected = sort.ConvertChildIterToIter (externalSelected);
+					foreach (TreeIter element in dictionaryStore [ViewModel.Selection.FirstOrDefault ()]) {
+						TreeIter externalSelected = element;
+						if (filter != null) {
+							externalSelected = filter.ConvertChildIterToIter (externalSelected);
+							externalSelected = sort.ConvertChildIterToIter (externalSelected);
+						}
+						Selection.SelectIter (externalSelected);
```
dfernandezramos: Won't this select always the last element even if you have selected the first one?

VAS.UI.Gtk2/UI/TreeView/TimelineEventsTreeView.cs
```diff
@@ -129,7 +129,9 @@ protected override void HandleTreeviewSelectionChanged (object sender, EventArgs
 				foreach (TimelineEventVM eventVM in GetSelectedViewModels ().OfType<TimelineEventVM> ()) {
 					eventVM.Selected = true;
 					events.Add (eventVM);
-					Selection.SelectIter (dictionaryStore [eventVM]);
+					foreach (TreeIter iter in dictionaryStore[eventVM]) {
+						Selection.SelectIter (iter);
```
dfernandezramos: Same here. Just asking if this won't select always the last item.

PR #257, opened 2017-04-25
LON-1023

SandraPG85:
Fix contextual menus from tree view and timeline

### DIFF
--- a/VAS.Drawing/Widgets/PlaysTimeline.cs
+++ b/VAS.Drawing/Widgets/PlaysTimeline.cs
@@ -285,18 +285,9 @@ namespace VAS.Drawing.Widgets
 
 		protected void ShowTimersMenu (Point coords)
 		{
-			if (PeriodsTimeline != null &&
-				coords.Y >= PeriodsTimeline.OffsetY &&
-				coords.Y < PeriodsTimeline.OffsetY + PeriodsTimeline.Height) {
-				TimerVM t = Selections.Select (p => (p.Drawable as TimerTimeNodeView).Timer).FirstOrDefault ();
-				if (ShowTimerMenuEvent != null) {
-					ShowTimerMenuEvent (t.Model, Utils.PosToTime (coords, SecondsPerPixel));
-				}
-			} else {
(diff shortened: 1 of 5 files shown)


VAS.Drawing/Widgets/PlaysTimeline.cs
```diff
@@ -285,18 +285,9 @@ protected virtual void RemoveEventTypeTimeline (EventTypeTimelineVM timelineVM)
 
 		protected void ShowTimersMenu (Point coords)
 		{
-			if (PeriodsTimeline != null &&
```
ylatuya: This a bit strange... the second path wasn't used in the old version?

VAS.UI.Gtk2/UI/Menu/MenuHelpers.cs
```diff
@@ -83,9 +83,13 @@ static public void FillAddToPlaylistMenu (MenuItem addToPlaylistMenu, IEnumerabl
 				};
 			}
 			item = new MenuItem (Catalog.GetString ("Create new playlist..."));
-			LicenseLimitationVM limitation = App.Current.LicenseLimitationsService.Get ("Presentations");
-			if (limitation != null) {
-				item.Sensitive = limitation.Count < limitation.Maximum;
+
+			// FIXME: Longomatch has not implemented the limitation service, remove the null check when it is done
```
ylatuya: Maybe fixed in https://github.com/fluendo/VAS/pull/221 already ?

VAS.Drawing/Widgets/PlaysTimeline.cs
SandraPG85: In the previous version the periodstimelineview do not exist. This has code has been extracted and moved to the LMPlaysTimeline

VAS.UI.Gtk2/UI/Menu/MenuHelpers.cs
SandraPG85: no, but it will be fixed in a future user story I added the check to allow menus appear in the meantime.

PR #261, opened 2017-05-03
Lon 959

vmartos:


VAS.Core/Interfaces/MVVMC/IVideoPlayerVM.cs
```diff
@@ -0,0 +1,27 @@
...
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using System;
+using VAS.Core.ViewModel;
+
+namespace VAS.Core.Interfaces.MVVMC
+{
+	public interface IVideoPlayerVM
+	{
+		VideoPlayerVM VideoPlayer { get; }
```
vguzmanp: This kind of interfaces aren't useful. They should be interfaces for the specific methods in VideoPlayerVM

VAS.Drawing/CanvasObjects/CanvasButtonObject.cs
```diff
@@ -79,17 +79,17 @@ public void Click (Point p, ButtonModifier modif)
...
 				clicked = true;
 			}
 		}
 
 		public override void ClickReleased ()
 		{
 			if (clicked) {
-				Active = !Active;
+				if (!Toggle) {
+					Active = !Active;
```
vguzmanp: What was happening here?

VAS.Services/Controller/TaggingController.cs
```diff
@@ -95,7 +95,7 @@ public override void Stop ()
 		public override void SetViewModel (IViewModel viewModel)
 		{
 			project = (ProjectVM)(viewModel as dynamic);
-			VideoPlayer = (VideoPlayerVM)(viewModel as dynamic);
+			VideoPlayer = (viewModel as IVideoPlayerVM)?.VideoPlayer;
```
vguzmanp: What will happen if this is null?

VAS.Services/ViewModel/DrawingToolVM.cs
```diff
@@ -16,22 +16,20 @@
...
 {
-	public class DrawingToolVM : DisposableBase, IViewModel
+	public class DrawingToolVM : ViewModelBase
 	{
 		/// <summary>
 		/// Gets or sets the project.
 		/// </summary>
 		/// <value>The project.</value>
-		public Project Project {
+		public virtual Project Project {
```
vmartos: because DrawingTool is not fully ported to MVVM

VAS.Core/Interfaces/MVVMC/IVideoPlayerVM.cs
vmartos: this types of interfaces are only used to cast and obtain the videoplayerVM in every controller, this is how was done in the VAS-46 task, maybe add a coment there if you find the name of the interface not suitable

VAS.Drawing/CanvasObjects/CanvasButtonObject.cs
vmartos: the canvas toggle button was not behaving correctly, because this first was changed only taking in consideration canvas button that weren't Toggle mode. Now for all types of canvas button the behaviour is correct

VAS.Services/Controller/TaggingController.cs
vmartos: Actually this is changed in VAS-46 task, there cannot be null and should throw an exception

VAS.Core/Interfaces/MVVMC/IVideoPlayerVM.cs
vguzmanp: Ok, so it's not an `IVideoPlayerVM`, it's more an `IVideoPlayerContainer` or something like that. But still I think that we don't win much like this. This structure is exactly the same as casting to `VideoPlayerVM` directly :S

VAS.Services/ViewModel/DrawingToolVM.cs
vguzmanp: Add a FIXME please

PR #262, opened 2017-05-03
Ra 167

vguzmanp:


VAS.Drawing/CanvasObjects/PinCanvasObject.cs
```diff
@@ -15,51 +15,88 @@
...
-			throw new NotImplementedException ();
+			Position = dst;
 		}
 
 		public override void Draw (IDrawingToolkit tk, Area area)
 		{
-			throw new NotImplementedException ();
+			if (!Visible) {
+				return;
+			}
```
SandraPG85: combine the if expression to return when needed

VAS.Drawing/CanvasObjects/PinCanvasObject.cs
```diff
@@ -15,51 +15,88 @@
...
-using System;
 using VAS.Core;
 using VAS.Core.Common;
 using VAS.Core.Interfaces.Drawing;
 using VAS.Core.Store.Drawables;
-using VAS.Drawing.CanvasObjects;
 
 namespace VAS.Drawing.CanvasObjects
 {
 	public class PinCanvasObject : FixedSizeCanvasObject, ICanvasSelectableObject
```
SandraPG85: add the proper summaries in the class and public methods...

VAS.UI.Gtk2/UI/Component/HotkeysConfigurationView.cs
```diff
@@ -115,7 +115,8 @@ void FillKeyConfigs (IEnumerable<KeyConfigVM> keyConfigs)
 				keyLabel.SetAlignment (0f, 0.5f);
 				keyLabel.WidthRequest = 200;
 				edit = new Button ();
-				edit.Bind (config.EditCommand);
+				//FIXME: This should use the BindingContext
```
SandraPG85: The binding context is being added, why not fix this now?

vguzmanp: I changed the ones that were straightforward, but this one wasn't, and it had nothing to do with my story.

VAS.Drawing/CanvasObjects/PinCanvasObject.cs
vguzmanp: Funny, I had the same comment from andoni a while ago:
https://bitbucket.org/fluendo/riftanalyst/pull-requests/130/dont-draw-mapobjects-if-their-viewmodel-is/diff#comment-28248561
My rationale here is the same: `One is the visibility, the other is the content`

SandraPG85: This can be arguable, since you can keep separate booleans and one unique conditional and point to way out that checks them. Also note that it is better to add a function if at the end there are more lines that checks nullability that actually logic in the function (I know that now is not the case but just to comment)

PR #263, opened 2017-05-04
Fix linking buttons

SandraPG85:


VAS.Core/Common/StateController.cs
```diff
@@ -484,11 +485,15 @@ NavigationState LastModalState ()
 
 	class NavigationState
 	{
-		public NavigationState (string name, IScreenState screenState)
+		NavigationState freezingState;
+		bool completeWhenUnload;
```
SandraPG85: In this case, variable are private so they must remain camelCase

VAS.Core/Store/DashboardButton.cs
```diff
@@ -111,7 +111,7 @@ public DashboardButton ()
 		/// <summary>
 		/// A list with all the outgoing links of this button
 		/// </summary>
-		public ObservableCollection<ActionLink> ActionLinks {
+		public RangeObservableCollection<ActionLink> ActionLinks {
```
SandraPG85: The view will work will the VM, but this class is the model, so which is you concern?

VAS.Core/ViewModel/DashboardButtonCollectionVM.cs
```diff
@@ -26,12 +28,42 @@ namespace VAS.Core.ViewModel
...
+			if (cachedButtonVMs.ContainsKey (model)) {
+				return cachedButtonVMs [model];
+			}
+
+			DashboardButtonVM buttonVM = base.CreateInstance (model);
+			foreach (ActionLinkVM link in buttonVM.ActionLinks) {
+				link.SourceButton = GetLinkedButton (link.Model.SourceButton);
+				link.DestinationButton = GetLinkedButton (link.Model.DestinationButton);
+			}
+
```
SandraPG85: In the function GetLinkedButton if the VM is not available it is created and added to the dictionary, so I don't understand what you mean.

VAS.Core/ViewModel/DashboardButtonVM.cs
```diff
@@ -174,13 +193,9 @@ public class DashboardButtonVM : ViewModelBase<DashboardButton>
...
-		public ObservableCollection<ActionLink> ActionLinks {
-			get {
-				return Model.ActionLinks;
-			}
-			set {
-				Model.ActionLinks = value;
-			}
+		public CollectionViewModel<ActionLink, ActionLinkVM> ActionLinks {
+			get;
+			set;
```
SandraPG85: when the model is set to the VM (DashboardButtonVM) it is also linked the model of the ActionLinks "CollectionViewModel"
public override DashboardButton Model {
			get {
				return base.Model;
			}
			set {
				base.Model = value;
				ActionLinks.Model = value.ActionLinks;
			}
		}

VAS.Core/Common/StateController.cs
```diff
@@ -484,11 +485,15 @@ NavigationState LastModalState ()
...
+
+		public NavigationState (string name, IScreenState screenState, bool completeWhenUnload = true)
 		{
 			Name = name;
 			ScreenState = screenState;
 			Completion = new TaskCompletionSource<bool> ();
+			this.completeWhenUnload = completeWhenUnload;
```
aameijide: You should change the local or class field name or it will understand that you're assigning it to the same field.

VAS.Core/Store/DashboardButton.cs
aameijide: Ok, I was confuse and I understood this file as the View.

PR #264, opened 2017-05-04
Use Interfaces instead of implicit cast operators

vmartos:


VAS.Core/Interfaces/MVVMC/IDashboardVM.cs
```diff
@@ -0,0 +1,27 @@
...
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using System;
+using VAS.Core.ViewModel;
+
```
SandraPG85: add a comment to the interfaces and the public properties, do this in the other added interfaces as well

VAS.Services/VideoPlayerController.cs
```diff
@@ -351,7 +351,10 @@ protected override void DisposeManagedResources ()
 
 		public override void SetViewModel (IViewModel viewModel)
 		{
-			playerVM = (VideoPlayerVM)(viewModel as dynamic);
+			playerVM = viewModel as VideoPlayerVM;
+			if (playerVM == null) {
+				playerVM = ((IVideoPlayerVM)viewModel).VideoPlayer;
```
SandraPG85: why are you setting the view model only if it is null ?

vguzmanp:
Coming from https://github.com/fluendo/VAS/pull/261#discussion_r114972128

I think this interface structure doesn't gives us much. When we talked about this I understood that the objects would contain an `IWhatever`, cast against that, and work with that.

I don't really understand what we win from doing 
```
IViewModel vm
(IWhateverVM)vm
vm.WhateverVM
```
vs
```
IViewModel vm
(WhateverVM)vm
```

vmartos:
The main problem here is that Implicit/explicit (cast) operators are deleted in favour of interfaces. Some controllers are shared between different states, making them receive different types of ViewModels. Example: ProjectAnalysisVM, ProjectManagerVM, etc. Then if we want to obtain the VideoPlayerVM we cannot do:
 (VideoPlayerVM) viewModel.
Since now there aren't any explicit operators to make the cast.
That's why the interfaces comes in place. If we know that the ViewModel implements the IVideoPlayerVM we can obtain the VideoPlayer.

Maybe we need to change all Interface names, instead of IVideoPlayerVM we can rename it to IVideoPlayerContainerVM or something more meaningful.

Using Interfaces cleans the code a little, because now the ViewModels don't have huge list of methods for just casting purposes, this reduces duplication of code, etc.

VAS.Services/VideoPlayerController.cs
vmartos: in this case one can do videoPlayerController.SetViewModel (videoPlayerVM), then i cast and check if the ViewModel is already of type VideoPlayerVM, If not then y cast to IVideoPlayerDealer to obtain the player

PR #269, opened 2017-05-10
Fix templates exporting

ylatuya:


VAS.Tests/Services/TestTemplatesController.cs
```diff
@@ -183,5 +185,96 @@ public void TestHandleSave_SaveStatic ()
...
+			team.List.Add (player);
+			TeamVM teamVM = new TeamVM { Model = team };
+			templatesController.ViewModel.ViewModels.Add (teamVM);
+			templatesController.ViewModel.Selection.Add (teamVM);
+			tempFile = Path.Combine (Path.GetTempPath (), Path.GetRandomFileName ());
+			tempFile = Path.ChangeExtension (tempFile, null);
+			using (var stream = File.AppendText (tempFile)) {
+				stream.WriteLine ("FOO");
+			}
+			mockDialogs.Setup (m => m.SaveFile (It.IsAny<string> (), It.IsAny<string> (),
```
dfernandezramos: Can't we do this mock setup just once in a TestFixtureSetup instead of setting it in each test?

VAS.Tests/Services/TestTemplatesController.cs
```diff
@@ -183,5 +185,96 @@ public void TestHandleSave_SaveStatic ()
...
+			DummyTeam team = new DummyTeam ();
+			team.List.Add (player);
+			TeamVM teamVM = new TeamVM { Model = team };
+			templatesController.ViewModel.ViewModels.Add (teamVM);
+			templatesController.ViewModel.Selection.Add (teamVM);
+			tempFile = Path.Combine (Path.GetTempPath (), Path.GetRandomFileName ());
+			tempFile = Path.ChangeExtension (tempFile, null);
+			using (var stream = File.AppendText (tempFile)) {
+				stream.WriteLine ("FOO");
+			}
```
dfernandezramos: We are doing this in more than one test. Can't we bring this to a single method and then call it each time is needed?

dfernandezramos:
👍

PR #270, opened 2017-05-10
Fix transitions when LoadState fails

ylatuya:
When LoadState fails in a transition we must make sure
that we restore the previous state.

### DIFF
--- a/VAS.Core/Common/StateController.cs
+++ b/VAS.Core/Common/StateController.cs
@@ -111,11 +111,23 @@ namespace VAS.Core
 				} else {
 					state = destination [transition] ();
 					if (!await state.LoadState (properties)) {
+						// If the transition failed and the stack is empty, load the home,
+						// otherwise show again the last state that was hidden at the start of the
+						// MoveTo
+						if (emptyStack) {
+							await PushNavigationState (home.Name, home.ScreenState);
+						} else {
+							if (!await lastState.Show ()) {
+								// This shouldn't fail... but just in case
+								Log.Error ("Last state couldn't be shown again, we'll move back home");
+								await PushNavigationState (home.Name, home.ScreenState);
+							}
+						}
 						return false;
 					}
 				}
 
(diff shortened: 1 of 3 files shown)


VAS.Core/Common/StateController.cs
```diff
@@ -111,11 +111,23 @@ public StateController ()
...
+						// otherwise show again the last state that was hidden at the start of the
+						// MoveTo
+						if (emptyStack) {
+							await PushNavigationState (home.Name, home.ScreenState);
+						} else {
+							if (!await lastState.Show ()) {
+								// This shouldn't fail... but just in case
+								Log.Error ("Last state couldn't be shown again, we'll move back home");
+							}
+							await PushNavigationState (home.Name, home.ScreenState);
```
vmartos: Shouldn't this be inside if (!await lastState.Show ()) {} ?

VAS.Core/Common/StateController.cs
```diff
@@ -556,17 +600,28 @@ public override bool Equals (object obj)
...
+			if (result) {
+				UpdateState (NavigationStateStatus.Frozen);
+			}
+			return result;
 		}
 
 		public async Task<bool> Hide ()
 		{
-			return await ScreenState.HideState ();
+			bool result = await ScreenState.HideState ();
```
vmartos: can we add a protection in every methood Hide,Unload, etc. that if CurrentState == The state you want to move returns a true?

VAS.Core/Common/StateController.cs
ylatuya: Yes ...

VAS.Core/Common/StateController.cs
ylatuya: It should be an error in the state machine.

vmartos: but then how can we solve errors when doing an MoveTo with emptystack, this is calling Hide to all already hidden states.

PR #271, opened 2017-05-10
Center dialogs on parent manually

ylatuya:
When the dialog's contructor has been called and Show ()
is called before setting TransientFor, we need to manually
center it.

### DIFF
--- a/VAS.UI.Gtk2/GUIToolkitBase.cs
+++ b/VAS.UI.Gtk2/GUIToolkitBase.cs
@@ -23,14 +23,12 @@ using Gtk;
 using VAS.Core;
 using VAS.Core.Common;
 using VAS.Core.Events;
-using VAS.Core.Filters;
 using VAS.Core.Interfaces;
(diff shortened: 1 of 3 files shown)


VAS.UI.Gtk2/GUIToolkitBase.cs
```diff
@@ -208,8 +206,10 @@ public Widget GetParentWidget (object parent)
 		protected void ShowModalWindow (IPanel panel, IPanel parent)
 		{
 			if (panel is Gtk.Dialog) {
-				(panel as Gtk.Dialog).TransientFor = ((Bin)parent).Toplevel as Window;
-				(panel as Gtk.Dialog).DeleteEvent += ModalWindowDeleteEvent;
+				var dialog = (panel as Gtk.Dialog);
+				dialog.TransientFor = ((Bin)parent).Toplevel as Window;
```
SandraPG85: Extract the cast to Dialog outside the if sentence and check the nullability in the if to avoid do it twice

PR #274, opened 2017-05-12
Fix exeption deleting buttons and selecting others

ylatuya:
This is due to the lack of syncrhonization of the Selection
after having removed the item

### DIFF
--- a/VAS.Core/ViewModel/TemplateViewModel.cs
+++ b/VAS.Core/ViewModel/TemplateViewModel.cs
@@ -15,11 +15,10 @@
 //  along with this program; if not, write to the Free Software
 //  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 //
-using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Specialized;
-using System.ComponentModel;
+using System.Linq;
 using VAS.Core.Common;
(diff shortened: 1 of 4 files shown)


VAS.Core/ViewModel/TemplateViewModel.cs
```diff
@@ -187,5 +186,20 @@ protected override void SyncLoadedModel ()
...
 		}
+
+		void HandleViewModelsChanged (object sender, NotifyCollectionChangedEventArgs e)
+		{
+			switch (e.Action) {
+			case NotifyCollectionChangedAction.Remove:
+				Selection.RemoveRange (Selection.Intersect (e.OldItems.OfType<TChildViewModel> ()));
+				break;
+			case NotifyCollectionChangedAction.Replace:
+				Selection.RemoveRange (Selection.Intersect (e.OldItems.OfType<TChildViewModel> ()));
```
vguzmanp: Why do you need to intersect it? oldItems should always be a subset of Selection, right?

ylatuya: OldItems can have elements that are not in Selection, RemoveRange will call Remove that will try to remove elements that are not in the Selection list.

vguzmanp: 👌

PR #279, opened 2017-05-15
Also sync all properties when the Model changes

ylatuya:
When the model is changed in a ViewModel we also want
the View to resync all properties, as if Sync() was called

### DIFF
--- a/VAS.Tests/MVVMC/TestViewModelBase.cs
+++ b/VAS.Tests/MVVMC/TestViewModelBase.cs
@@ -170,6 +170,20 @@ namespace VAS.Tests.MVVMC
 
 			Assert.IsNull (propertyName);
 		}
+
+		[Test]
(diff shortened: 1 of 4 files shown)


VAS.Tests/MVVMC/TestViewModelBase.cs
```diff
@@ -170,6 +170,20 @@ public void TestSync ()
...
+		{
+			bool called = false;
+			var viewModel = new ViewModelBase<StorableBase> ();
+			viewModel.PropertyChanged += (sender, e) => {
+				called |= e.PropertyName == null;
+			};
+
+			viewModel.Model = new StorableBase ();
+
+			Assert.IsTrue (called);
```
vguzmanp: Can you add a test for the case of an overridden Model?

PR #280, opened 2017-05-15
Use Xamarin.Mac from the prefix

ylatuya:
This should prevent xbuild to copy the custom OS X
assemblies referenced from Xamarin.Mac, like System.Core,
to the bin folder. We prevent this way mixing framework versions
since those could have been compiled against a more recent
framework than the one we use.


vmartos:
👍

PR #281, opened 2017-05-15
LON-1046: Project events are drawed multiline

aameijide:


VAS.Core/Interfaces/Drawing/IDrawingToolkit.cs
```diff
@@ -87,21 +87,22 @@ public interface IDrawingToolkit
...
+		void DrawArea (params Point [] vertices);
 
 		void DrawPoint (Point point);
 
 		void DrawCircle (Point center, double radius);
 
 		void DrawEllipse (Point center, double axisX, double axisY);
 
-		void DrawText (Point point, double width, double height, string text, bool escape = false, bool ellipsize = false);
+		void DrawText (Point point, double width, double height, string text, bool escape = false,
```
ylatuya: I don't think this is fixing the actual problem which is that the specified height is smaller than it should be for the font size. As the pango documentation says, a negative height only affects the number of lines per paragraph but it does not limit the total height of the layout. Using this API a user could set a a height and the text would go beyong that desired height.

aameijide: Ok, I'll check that again. The strange thing here is that if you set it positive, with the correct height, only writes the first line, maybe can we do like this?
`if(!multiline) { layout.SetPangoLayoutHeight (layoutHeight); }`

aameijide:
From chat:
Andoni Morales [10:48] 
@aameijide  al final era solo convertir el height en pixeles a pango units no ?

Adrià Ameijide [10:50] 
@ylatuya si

Andoni Morales [10:51] 
+1

PR #285, opened 2017-05-19
Fix seeks when a timer or period changes

ylatuya:
Also cleanup the controller a bit

### DIFF
--- a/VAS.Core/MVVMC/Command.cs
+++ b/VAS.Core/MVVMC/Command.cs
@@ -23,36 +23,27 @@ using VAS.Core.Common;
 
 namespace VAS.Core.MVVMC
 {
+	/// <summary>
+	/// Command implementation of <see cref="ICommand"/>.
+	/// </summary>
 	public class Command : ICommand
 	{
 		public event EventHandler CanExecuteChanged;
 		protected Func<object, bool> canExecute;
-		readonly Func<object, Task> execute;
+		protected Func<object, Task> execute;
 		bool executable;
 		bool isExecuting;
 
-		public Command (Func<object, Task> execute)
+		protected Command ()
 		{
-			Contract.Requires (execute != null);
-
-			this.execute = execute;
-			Executable = true;
 		}
 
 		public Command (Action<object> execute)
 		{
 			Contract.Requires (execute != null);
 
-			this.execute = (o) => {
-				execute (o);
-				return AsyncHelpers.Return ();
-			};
+			this.execute = (o) => { execute (o); return AsyncHelpers.Return (); };
 			Executable = true;
-		}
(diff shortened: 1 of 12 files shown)


VAS.Services/Controller/EventsController.cs
```diff
@@ -115,14 +96,22 @@ public override void Stop ()
...
 		}
 
 		public override void SetViewModel (IViewModel viewModel)
 		{
-			PlayerVM = ((IVideoPlayerDealer)viewModel).VideoPlayer;
+			VideoPlayer = ((IVideoPlayerDealer)viewModel).VideoPlayer;
 			Timeline = ((ITimelineDealer)viewModel).Timeline;
-			projectVM = (viewModel as IProjectDealer)?.Project;
-			capturer = (viewModel as ICapturerBinDealer)?.Capturer;
+			Project = (viewModel as IProjectDealer)?.Project;
```
aameijide: Shouldn't we cast here with `((IProjectDealer)viewmodel))`?

VAS.Core/MVVMC/Command.cs
```diff
@@ -27,34 +27,20 @@ public class Command : ICommand
 	{
 		public event EventHandler CanExecuteChanged;
 		protected Func<object, bool> canExecute;
-		readonly Func<object, Task> execute;
+		protected Func<object, Task> execute;
 		bool executable;
 		bool isExecuting;
 
-		public Command (Func<object, Task> execute)
+		public Command ()
```
ylatuya: It's needed by AsyncCommand, but it should protected, not public

VAS.Services/Controller/EventsController.cs
ylatuya: the viewModel might not be an IProjectDealer, so the cast must be optional

VAS.Core/MVVMC/Command.cs
```diff
@@ -71,34 +57,13 @@ public Command (Action<object> execute, Func<object, bool> canExecute) : this (e
 		/// <summary>
 		/// Initializes a new instance of the <see cref="T:VAS.Core.MVVMC.Command"/> class with an async function.
```
ylatuya: I have remove all constructors summaries and documented the class, since the constructors are quite redundant

VAS.Core/MVVMC/Command.cs
```diff
@@ -23,38 +23,29 @@
 
 namespace VAS.Core.MVVMC
 {
+	/// <summary>
+	/// Command implementationn of <see cref="ICommand"/>.
```
aameijide: typo: implementatio*nn*

PR #287, opened 2017-05-22
Ra 845

vguzmanp:


VAS.UI.Gtk2/GUIToolkitBase.cs
```diff
@@ -183,11 +184,33 @@ public HotKey SelectHotkey (HotKey hotkey, object parent = null)
...
 		public void Invoke (EventHandler handler)
 		{
 			Application.Invoke (handler);
 		}
 
+		/// <summary>
+		/// Queues the handler in the main thread synchronously. It waits until the handler is executed.
+		/// </summary>
+		/// <param name="handler">Handler.</param>
+		public void InvokeSync (EventHandler handler)
```
ylatuya: Overall, I think we can make this function much more async friendly now that we are adding new API. For example:

`
	public Task<T> Invoke<T> (this Func<Task<T>> handler)
		{
			TaskCompletionSource<T> tcs = new TaskCompletionSource<T> ();
			Application.Invoke (async (sender, e) => {
				var res = await handler ();
				tcs.SetResult (res);
			});
			return tcs.Task;
		}
`

VAS.Services/AppUpdater/SparkleWin.cs
```diff
@@ -34,12 +41,29 @@ public class SparkleWin : IAppUpdater
...
+
+		CanShutdownCallback can_shutdown_callback;
+		ShutdownRequestCallback shutdown_request_callback;
 
 		public void Start (string companyName, string appName, string version, string feedURL, string libDir)
 		{
 			win_sparkle_set_automatic_check_for_updates (1);
 			win_sparkle_set_app_details (companyName, appName, version);
 			win_sparkle_set_appcast_url (feedURL);
+			Log.Information ("Registering win_sparkle_set_can_shutdown_callback");
```
vguzmanp: 👌

VAS.Services/AppUpdater/SparkleWin.cs
```diff
@@ -52,5 +76,32 @@ public void CheckForUpdates ()
...
+				task.Wait ();
+				result = task.Result;
+			});
+			return result;
+		}
+
+		void HandleShutdownRequestCallback ()
+		{
+			Log.Information ("WinSparkle ShutdownRequestCallback called");
+			App.Current.GUIToolkit.Invoke ((sender, e) => {
```
vguzmanp: From the winsparkle docs (https://github.com/vslavik/winsparkle/blob/master/include/winsparkle.h#L321):

> 
> @note There's no guarantee about the thread from which the callback is called,
>           except that it certainly *won't* be called from the app's main thread.
>           Make sure the callback is thread-safe.

VAS.Services/AppUpdater/SparkleWin.cs
```diff
@@ -52,5 +76,29 @@ public void CheckForUpdates ()
...
+			// Because this is a delegate and we don't have an option to await the result of the task, Wait() it with a
+			// continuation in a different thread to avoid blocking.
+			return taskResult.Result;
+		}
+
+		void HandleShutdownRequestCallback ()
+		{
+			Log.Information ("WinSparkle ShutdownRequestCallback called");
+			App.Current.GUIToolkit.Invoke ((sender, e) => {
+				Task task = App.Current.GUIToolkit.Quit ();
```
ylatuya: conver to to async (sender, e) await App.Current.GUIToolkit.Quit (); or just call Quit(), you will be in the main thread :)

VAS.UI.Gtk2/GUIToolkitBase.cs
```diff
@@ -183,11 +184,39 @@ public HotKey SelectHotkey (HotKey hotkey, object parent = null)
...
 		}
 
+		/// <summary>
+		/// Invokes the handler in the main thread asynchronously.
+		/// </summary>
+		/// <param name="handler">Handler.</param>
+		public async Task<T> Invoke<T> (Func<Task<T>> handler)
+		{
+			if (App.IsMainThread) {
+				Log.Warning ("Invoke called from the main thread");
```
ylatuya: If this is a warning here, it should also be a warning in the other function.

VAS.Services/AppUpdater/SparkleWin.cs
ylatuya: This change is still missing

vguzmanp: I changed it now. I changed it to be the same as the other callback, to make sure we wait until the end of the `Quit` call to notify sparkle.

VAS.UI.Gtk2/GUIToolkitBase.cs
vguzmanp: I added a warning to the other, but the code is still the same.

PR #288, opened 2017-05-22
Fix timerule seeks when an event is loaded

ylatuya:


VAS.Drawing/Widgets/Timerule.cs
```diff
@@ -57,6 +57,7 @@ public Timerule (IWidget widget) : base (widget)
 			BackgroundColor = App.Current.Style.PaletteBackgroundDark;
 			Accuracy = 5.0f;
 			PlayerMode = false;
+			UseAbsoluteDuration = false;
```
vguzmanp: Wasn't this already working before, without the new logic? I expected this to be something that the VM is sending differently, with an easy fix in the VM.

VAS.Drawing/Widgets/Timerule.cs
```diff
@@ -307,7 +312,13 @@ public override void Draw (IContext context, Area area)
 		protected override void StartMove (Selection sel)
 		{
 			WasPlaying = ViewModel.Playing;
-			ViewModel.LoadEvent (null, false);
+			if (UseAbsoluteDuration) {
+				// Seeks can happen outside the region of the loaded element,
+				// so we have to unload it
+				viewModel.LoadEvent (null, false);
```
vguzmanp: Shouldn't we unload it when the seek goes out of the event, not always?

vguzmanp: Use `ViewModel` instead of `viewModel`

VAS.Drawing/Widgets/Timerule.cs
```diff
@@ -360,8 +371,16 @@ protected override void HandleDoubleClick (Point coords, ButtonModifier modif)
 		void HandlePropertyChangedEventHandler (object sender, PropertyChangedEventArgs e)
 		{
 			if (e.PropertyName == nameof (VideoPlayerVM.Duration)) {
-				Duration = ViewModel.Duration;
-			} else if (AutoUpdate && e.PropertyName == nameof (VideoPlayerVM.CurrentTime)) {
+				if (!UseAbsoluteDuration) {
+					Duration = ViewModel.Duration;
+				}
+			}
+			if (e.PropertyName == nameof (VideoPlayerVM.AbsoluteDuration)) {
```
vguzmanp: Why not use `NeedsSync`?

VAS.Drawing/Widgets/Timerule.cs
```diff
@@ -360,8 +371,16 @@ protected override void HandleDoubleClick (Point coords, ButtonModifier modif)
 		void HandlePropertyChangedEventHandler (object sender, PropertyChangedEventArgs e)
 		{
 			if (e.PropertyName == nameof (VideoPlayerVM.Duration)) {
-				Duration = ViewModel.Duration;
-			} else if (AutoUpdate && e.PropertyName == nameof (VideoPlayerVM.CurrentTime)) {
+				if (!UseAbsoluteDuration) {
```
vguzmanp: Merge this into the same if (`if (e.PropertyName == nameof (VideoPlayerVM.Duration))`)

VAS.Tests/Drawing/Widgets/TestTimerule.cs
```diff
@@ -59,5 +66,30 @@ public void TestAutoUpdateDisabled ()
...
+		public void StartSeek_NotUserAbsoluteDuration_EventNotUnloaded ()
+		{
+			timerule.UseAbsoluteDuration = false;
+
+			widgetMock.Raise ((obj) => obj.ButtonPressEvent += null,
+							  new Point (0, 0), (uint)0, ButtonType.Left, ButtonModifier.None,
+							  ButtonRepetition.Single);
+
+			videoPlayerMock.Verify (p => p.Pause (false), Times.Once ());
+			videoPlayerMock.Verify (p => p.UnloadCurrentEvent (), Times.Never ());
```
vguzmanp: You shouldn't be able to seek outside `duration` if `UseAbsoluteDuration` is false, right? Can you add a test for that?

VAS.Drawing/Widgets/Timerule.cs
ylatuya: This wasn't working, we are using the same timerule object for 2 different scenarios, the player in which the timerule changes dynamically and the timeline, in which the timeline alaways shows the full duration and there a single VM which is the PlayerVM

VAS.Drawing/Widgets/Timerule.cs
ylatuya: This is how it was working before, any seeks in the timeline unloads the event.

VAS.Tests/Drawing/Widgets/TestTimerule.cs
ylatuya: This limitation is only in the View itself, only displaying visually the seekable region.

vguzmanp: So, with `!UseAbsoluteDuration`, if there is an event loaded that starts at 10, and you click on `0,0`, we are seeking to 10. And with `UseAbsoluteDuration`, with the same event loaded and the same click, we seek to 0.

I think it's an easy and useful test.

ylatuya: But that's handled by the controller, not the view. The view always seeks to 1,0 and the controller changes that seek into either 1,0 or event.starttime + 1,0

vguzmanp: True

vguzmanp:
Squash the fixups and 👍

PR #289, opened 2017-05-26
Fix commands constructors and add unit tests

ylatuya:


VAS.Core/MVVMC/Command.cs
```diff
@@ -219,21 +229,25 @@ public Command (Action<T> execute, Func<T, bool> canExecute) : base (o => execut
 	/// </summary>
 	public class AsyncCommand<T> : Command<T>
 	{
-		public AsyncCommand (Func<T, Task> execute) : base (o => execute (o))
+		public AsyncCommand (Func<T, Task> execute)
 		{
-			Contract.Requires (execute != null);
+			this.execute = o => execute ((T)o);
```
vguzmanp: Shouldn't this keep the contract.Requires?

VAS.Tests/MVVMC/TestCommand.cs
```diff
@@ -31,5 +31,70 @@ public void TestCommandInitExecutable ()
...
+
+			Assert.IsTrue (command.CanExecute (true));
+		}
+
+		[Test]
+		public void CanExecute_CommandParametrizedCanExecute_OK ()
+		{
+			var command = new Command<bool> (o => { }, o => o);
+
+			Assert.IsTrue (command.CanExecute (true));
```
vguzmanp: Didn't we need to call `EmitCanExecute` or something like that to update the CanExecute? I mean, if you pass `false` here, as you haven't called the emit, will it return `false`?

ylatuya: That's not how it works, If a funcion is passed in the constructor, CanExecute will always call that function , in this case o => o, or p => !projectsManager.Selection.Any() for example.
EmitCanExecute is just to notify views that the command execution state might have changed, so they should call again CanExecute to check the current state.

VAS.Core/MVVMC/Command.cs
ylatuya: yes, already fixed but not pushed

PR #290, opened 2017-05-29
RA-714

dfernandezramos:
RA-714

### DIFF
--- a/VAS.UI.Helpers.Gtk2/Bindings/ImageBinding.cs
+++ b/VAS.UI.Helpers.Gtk2/Bindings/ImageBinding.cs
@@ -50,9 +50,9 @@ namespace VAS.UI.Helpers.Bindings
 		protected override void WriteViewValue (Image val)
 		{
 			if (width != 0 && height != 0) {
-				image.Pixbuf = val.Scale (width, height).Value;
+				image.Pixbuf = val?.Scale (width, height).Value;
 			} else {
-				image.Pixbuf = val.Value;
+				image.Pixbuf = val?.Value;
 			}
 		}
 	}
--- a/VAS.UI.Helpers.Gtk2/Bindings/BindingsExtensions.cs
+++ b/VAS.UI.Helpers.Gtk2/Bindings/BindingsExtensions.cs
@@ -163,5 +163,23 @@ namespace VAS.UI.Helpers.Bindings
 		{
 			return new ColorButtonBinding (colorButton, propertyExpression);
 		}
+
+		/// <summary>
+		/// Bind the specified spinner to a property by name.
+		/// </summary>
+		/// <param name="spinner">Spinner.</param>
+		public static SpinnerBinding Bind (this Gtk.Image spinner, Expression<Func<IViewModel, bool>> propertyExpression)
+		{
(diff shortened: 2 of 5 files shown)


vguzmanp:
Wrong commit message

VAS.UI.Helpers.Gtk2/Bindings/ImageBinding.cs
```diff
@@ -50,9 +50,9 @@ protected override void UnbindView ()
 		protected override void WriteViewValue (Image val)
 		{
 			if (width != 0 && height != 0) {
-				image.Pixbuf = val.Scale (width, height).Value;
+				image.Pixbuf = val?.Scale (width, height).Value;
 			} else {
-				image.Pixbuf = val.Value;
```
vguzmanp: In what case do we set the image to null?

It would be better to have a [guard clause](https://fluendo.atlassian.net/wiki/display/Team/Coding+Style#CodingStyle-Guardclauses) for this

dfernandezramos: I have seen this with Andoni when doing `viewModel.Sync()`. If the image property of the viewmodel is not set at this poing we will get a crash because this method is not handling null values.

VAS.UI.Helpers.Gtk2/Bindings/ImageBinding.cs
```diff
@@ -49,10 +49,12 @@ protected override void UnbindView ()
 
 		protected override void WriteViewValue (Image val)
 		{
-			if (width != 0 && height != 0) {
-				image.Pixbuf = val.Scale (width, height).Value;
-			} else {
-				image.Pixbuf = val.Value;
+			if (val != null) {
```
vguzmanp: Hm... now it's ignoring the WriteValue if the image is null. Your previous code assigned it to null. Which one do we want?

dfernandezramos: ¯\_(ツ)_/¯ both work for me. So.........

VAS.Core/MVVMC/ViewModelBase.cs
```diff
@@ -138,6 +138,10 @@ protected override void ForwardPropertyChanged (object sender, PropertyChangedEv
 			Log.Verbose ($"ViewModelBase<T> {this} - Forwarding PropertyName: {e.PropertyName} with sender: {sender}");
 			base.ForwardPropertyChanged (sender, e);
 		}
+
+		protected virtual void HandlePropertyChanged (object sender, System.ComponentModel.PropertyChangedEventArgs e)
```
vguzmanp: Did you use this in the end?

dfernandezramos: Yes. Another thing is if I had used correctly. But yes

VAS.UI.Helpers.Gtk2/Bindings/ImageBinding.cs
vguzmanp: poor guy, it's missing an arm, here, have mine `\¯\_(ツ)`

PR #302, opened 2017-06-01
Ra 883

vguzmanp:


VAS.Drawing/SelectionCanvas.cs
```diff
@@ -317,7 +317,7 @@ protected virtual Selection GetSelection (Point coords, bool inMotion = false, b
 
 			/* Iterate over all the objects now */
 			if (sel == null) {
-				foreach (ICanvasSelectableObject co in Objects.OfType<ICanvasSelectableObject> ()) {
+				foreach (ICanvasSelectableObject co in Objects.OfType<ICanvasSelectableObject> ().Reverse ()) {
```
ylatuya: I would extend the comments to explain why we do it.

VAS.Drawing/Widgets/Blackboard.cs
```diff
@@ -297,6 +297,15 @@ public void Zoom (double zoom, Point center = null)
 			currentZoom = zoom;
 		}
 
+		protected override void DrawObjects (Area area)
```
ylatuya: Add comments to explain why we want this behaviour

VAS.Drawing/Widgets/Blackboard.cs
```diff
@@ -554,6 +563,40 @@ public void MoveSelected (Point offset)
 			}
 			widget.ReDraw ();
 		}
+
+		public void MoveToBack ()
```
ylatuya: With multiple selection I am not sure how that will work, but will see later :)

VAS.UI.Gtk2/UI/Dialog/DrawingTool.cs
```diff
@@ -322,6 +322,18 @@ public KeyContext GetKeyContext ()
 					() => blackboard.MoveSelected (new Point (0, MOVE_OFFSET)),
 					continueChain: false)
 			);
+			keyContext.AddAction (
+				new KeyAction (
+					App.Current.HotkeysService.GetByName (DrawingToolHotkeys.DRAWING_TOOL_MOVE_TO_FRONT),
+					() => blackboard.MoveToFront (),
```
ylatuya: continueChain is false by default already

VAS.Drawing/Widgets/Blackboard.cs
vguzmanp: Added support for multiselection, it's keeping the selection order instead of the drawing order, buuut it "works"

PR #304, opened 2017-06-01
Add support for watermarks

ylatuya:


VAS.Core/Common/Watermark.cs
```diff
@@ -0,0 +1,97 @@
...
+		/// The offset y, normalized from 0 to 1 based on output video height
+		/// </summary>
+		public double OffsetY;
+
+		/// <summary>
+		/// Gets the default watermark.
+		/// </summary>
+		/// <returns>The default watermark.</returns>
+		/// <param name="position">Position.</param>
+		public static Watermark GetDefaultWatermark (WatermarkPosition position, VideoStandard videoStandard)
```
vguzmanp: Missing tests

system_uuid/.deps/vas-system-uuid.Po
```diff
@@ -0,0 +1 @@
+# dummy
```
vguzmanp: ?

VAS.Core/Common/Watermark.cs
vguzmanp: I don't like this method's name. It's not getting the default watermark, it's configuring many things, calculating the offsets and creating the watermark.

VAS.UI.Gtk2/UI/Dialog/VideoEditionProperties.cs
```diff
@@ -56,8 +56,14 @@ public VideoEditionProperties (Window parent)
 				OutputDir = mediafilechooser2.CurrentPath;
 			};
 
-			watermarkcheckbutton.Active = true;
-			watermarkcheckbutton.Sensitive = !App.Current.LicenseManager.LicenseStatus.Limited;
+			watermarkcheckbutton.Active = App.Current.Config.AddWatermark;
+			//FIXME: This Logic should be changed once Longomatch initializes LicenseManager
+			if (App.Current.LicenseManager != null) {
```
vguzmanp: If it is null, we should set it to true, or fail, shouldn't we?

VAS.Core/Common/Watermark.cs
```diff
@@ -0,0 +1,97 @@
...
+
+		/// <summary>
+		/// The offset x, normalized from 0 to 1 based on output video height
+		/// </summary>
+		public double OffsetX;
+
+		/// <summary>
+		/// The offset y, normalized from 0 to 1 based on output video height
+		/// </summary>
+		public double OffsetY;
```
vguzmanp: Make these properties

VAS.UI.Gtk2/UI/Dialog/VideoEditionProperties.cs
```diff
@@ -125,6 +134,12 @@ protected virtual void OnButtonOkClicked (object sender, System.EventArgs e)
 			encSettings.EnableAudio = audiocheckbutton.Active;
 			encSettings.EnableTitle = descriptioncheckbutton.Active;
 
+			if (watermarkcheckbutton.Active) {
+				encSettings.Watermark = Watermark.GetDefaultWatermark (WatermarkPosition.TOP_RIGHT, encSettings.VideoStandard);
+			} else {
+				encSettings.Watermark = null;
+			}
```
vguzmanp: Why are all the settings not configured if there's not a watermark?

VAS.Core/StyleConf.cs
```diff
@@ -241,6 +241,10 @@ public class StyleConf
 		public const string PrelightEye = "images/events/el_eye_p" + IMAGE_EXT;
 		public const string InsensitiveEye = "images/events/el_eye_i" + IMAGE_EXT;
 
+		//Watermark
+		public const int WatermarkPadding = 20;
+		public const double WatermarkHeightNormalization = 0.087;
```
vmartos: Is the Height of the Watermark normalized from 0 to 1, the story says that always it should be 8,7% from the total Height of the video

VAS.UI.Gtk2/UI/Dialog/VideoEditionProperties.cs
vmartos: Longomatch can configure the watermark in preferences, that's why i do:

watermarkcheckbutton.Active = App.Current.Config.AddWatermark;

First. 

And I added a FIXME to change it in the future when LongoMatch should have always the LicenseManager initialized, if not it should throw.

VAS.Tests/Utils.cs
```diff
@@ -266,6 +267,44 @@ public DummyDashboardManagerVM (DashboardVM vm)
...
+		}
+
+		public Image LoadImage (string name, int width = 0, int height = 0)
+		{
+			return GetDummyImage ();
+		}
+
+		public void Register (Assembly assembly)
+		{
+			throw new NotImplementedException ();
```
vguzmanp: Do you really want to throw here?

VAS.Tests/Core/Common/TestWatermark.cs
```diff
@@ -0,0 +1,92 @@
...
+using NUnit.Framework;
+using VAS.Core.Common;
+
+namespace VAS.Tests.Core.Common
+{
+	[TestFixture]
+	public class TestWatermark
+	{
+		[Test]
+		public void TestWatermarkPosition_TopLeft ()
```
vguzmanp: Tests don't need to start with `Test`

VAS.Tests/Core/Common/TestWatermark.cs
```diff
@@ -0,0 +1,92 @@
...
+
+			Assert.AreEqual (expectedOffsetX, watermark.OffsetX);
+			Assert.AreEqual (expectedOffsetY, watermark.OffsetY);
+			Assert.AreEqual (expectedHeight, watermark.Image.Height);
+		}
+
+		[Test]
+		public void TestWatermarkPosition_BottomLeft ()
+		{
+			VideoStandard videoStandard = VideoStandards.P480;
```
vguzmanp: Why do you change the videoStandard in each test?

vmartos: Just to ensure that the offsets and image height is correctly calculated with different video sizes inside the ConfigureNewWatermark function.

vguzmanp: So... those are different tests, for the position, and for the standard

VAS.Tests/Core/Common/TestWatermark.cs
```diff
@@ -25,7 +25,7 @@ namespace VAS.Tests.Core.Common
 	public class TestWatermark
 	{
 		[Test]
-		public void ConfigureWatermark_PositionTopLeft ()
+		public void ConfigureWatermark_PositionTopLeft_VideoStandard_P720 ()
```
vguzmanp: The third part of the test should be the expected result.

https://fluendo.atlassian.net/wiki/display/APP/Coding+Style#CodingStyle-Testnaming

VAS.Tests/Core/Common/TestWatermark.cs
vguzmanp: You renamed the tests, but I meant that they should be different tests.

Ideally, all the combinations between videostandard & position. As that is quite redundant, at least set one of the two variables, and test with all the values of the other variable.

For example:
```
P480-Top_left
P480-Top_right
P480-Bot_left
P480-Bot_right

Top_left-P240
Top_left-P480
Top_left-P720
Top_left-P1080
```

PR #305, opened 2017-06-08
Ra 720

aameijide:


VAS.Core/Store/Series.cs
```diff
@@ -0,0 +1,64 @@
...
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using System;
+using VAS.Core.Common;
+using VAS.Core.MVVMC;
+
+namespace VAS.Core.Store
+{
+	public class Series : BindableBase
+	{
```
SandraPG85: if it is not storable i think that should be in another place.

VAS.Core/Store/Series.cs
```diff
@@ -0,0 +1,64 @@
...
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using System;
+using VAS.Core.Common;
+using VAS.Core.MVVMC;
+
+namespace VAS.Core.Store
+{
```
SandraPG85: add a summary for the class

VAS.Core/ViewModel/SeriesVM.cs
```diff
@@ -0,0 +1,89 @@
...
+using VAS.Core.Common;
+using VAS.Core.Interfaces.MVVMC;
+using VAS.Core.MVVMC;
+using VAS.Core.Store;
+
+namespace VAS.Core.ViewModel
+{
+	public class SeriesVM : BindableBase, IViewModel<Series>
+	{
+		public SeriesVM ()
```
SandraPG85: in this case I don't think that we need a model, you receive data (model) and paint it in a series chart but the series itself is not a model to me.

PR #312, opened 2017-06-19
RA-899

dfernandezramos:


VAS.Core/Common/Utils.cs
```diff
@@ -173,6 +173,23 @@ public static Stream GetEmbeddedResourceFileStream (string resourceId)
 		}
 
 		/// <summary>
+		/// Gets the embedded resource text file.
+		/// </summary>
+		/// <returns>The embedded resource text file.</returns>
+		/// <param name="resourceId">Resource identifier.</param>
+		public static string GetEmbeddedResourceTextFile (string resourceId)
```
ylatuya: Use the new ResourcesLocator instead.

PR #323, opened 2017-06-28
RA-971

dfernandezramos:


VAS.Core/Common/StateController.cs
```diff
@@ -405,7 +405,8 @@ public bool UnRegister (string transition)
 				return false;
 			}
 			await App.Current.EventsBroker.Publish (new NavigationEvent { Name = state.Name });
-			return await App.Current.Navigation.Push (state.ScreenState.Panel);
+			var result = await App.Current.Navigation.Push (state.ScreenState.Panel);
+			return result;
```
vguzmanp: WTF? What's the difference?

ylatuya: This shouldn't be merged, it seems a leftover from some tests to add a breakpoint

PR #325, opened 2017-07-04
Ra 178

vmartos:


VAS.Core/Interfaces/IPlaylistElement.cs
```diff
@@ -34,7 +34,7 @@ public interface IPlaylistElement : IChanged, INotifyPropertyChanged
 			get;
 		}
 
-		bool Selected {
```
vmartos: It was not moved, it was changed to Playing, I cannot remove from the Model because the VideoPlayerController stills works with Models.

VAS.Core/Interfaces/GUI/IDialogs.cs
```diff
@@ -67,7 +67,7 @@ public interface IDialogs
 		/// </summary>
 		/// <returns>The file.</returns>
 		/// <param name="parent">Parent widget.</param>
-		MediaFile OpenMediaFile (object parent);
+		MediaFile OpenMediaFile (object parent = null);
```
vmartos: We always use this method by calling OpenMediaFile(null), this just sets a default value to just call OpenMediaFile()

VAS.Core/MVVMC/LimitedCollectionViewModel.cs
```diff
@@ -128,7 +134,7 @@ void FillLimitedViewModels (IEnumerable<TViewModel> viewmodels)
 				viewmodels = viewmodels.Take (Limitation.Maximum);
 			}
 
-			LimitedViewModels.Replace (viewmodels);
+			LimitedViewModels?.Replace (viewmodels);
```
vmartos: Yep, In the Constructor of the base class (CollectionViewModel) is setting the Model: Model = new RangeObservableCollection<TModel> (); but at this point LimitedViewModels is null. I can add a protection though on:

protected override void SetModel (RangeObservableCollection<TModel> model)
 +		{
 +			base.SetModel (model);
if(LimitedViewModels != null)
 +			FillLimitedViewModels (base.ViewModels);
 +		}

VAS.Core/ViewModel/MenuNodeVM.cs
```diff
@@ -0,0 +1,67 @@
...
+	/// <summary>
+	/// Menu node vm, can be configured as a MenuNode or as a SubMenu
+	/// </summary>
+	public class MenuNodeVM : ViewModelBase
+	{
+		/// <summary>
+		/// Initializes a new instance of the <see cref="T:VAS.Core.ViewModel.MenuNodeVM"/> as a MenuNode with a command.
+		/// </summary>
+		/// <param name="command">Command.</param>
+		public MenuNodeVM (Command command, object commandParameter = null, string name = null)
```
vmartos: Command only accept one parameter

VAS.Core/ViewModel/PlaylistCollectionVM.cs
```diff
@@ -87,6 +98,80 @@ public PlaylistCollectionVM ()
...
+		}
+
+		[PropertyChanged.DoNotNotify]
+		public Command InsertVideoCommand {
+			get;
+			protected set;
+		}
+
+		[PropertyChanged.DoNotNotify]
+		public AsyncCommand<PlaylistPosition> InsertImageCommand {
```
vmartos: InsertImage is a specific case where it needs to be asynchronous, because the call to App.Current.Dialog.OpenImage is asynchronous.

VAS.Core/ViewModel/PlaylistElementVM.cs
```diff
@@ -68,4 +73,37 @@ public class PlaylistElementVM : ViewModelBase<IPlaylistElement>
...
+	public class PlaylistVideoVM : PlaylistElementVM
+	{
+	}
+
+	public class PlaylistImageVM : PlaylistElementVM
+	{
+	}
+
+	public class PlaylistDrawingVM : PlaylistElementVM
+	{
```
vmartos: The Models are different but they implement IPlaylistElement, since PlaylistElementVM has as a model IPlaylistElement it should work. But we need different ViewModels to know the correct ViewModel type for rendering. If someone needs to access specific Model properties it should override the Model from the PlaylistElementVM to access them

VAS.Services/Controller/PlaylistController.cs
```diff
@@ -224,6 +235,53 @@ void HandleRenderPlaylist (RenderPlaylistEvent e)
...
+			if (playlist != null) {
+				Save (playlist.Model);
+			}
+		}
+
+		void InsertPlaylistElements (IPlaylistElement element, PlaylistPosition position)
+		{
+			foreach (var playlist in playlistViewModel.LimitedViewModels) {
+				bool save = false;
+				foreach (var ple in playlist.Selection) {
```
vmartos: Yes, Is a new feature that longomatch users was requesting, for example inserting the same image or video before some elements as an introduction.

VAS.Services/Controller/PlaylistController.cs
```diff
@@ -224,6 +235,53 @@ void HandleRenderPlaylist (RenderPlaylistEvent e)
...
+			await App.Current.StateController.MoveToModal (EditPlaylistElementState.NAME, properties, true);
+			var playlist = playlistViewModel.ViewModels.FirstOrDefault ((arg) => arg.ViewModels.Contains (e.Object));
+			if (playlist != null) {
+				Save (playlist.Model);
+			}
+		}
+
+		void InsertPlaylistElements (IPlaylistElement element, PlaylistPosition position)
+		{
+			foreach (var playlist in playlistViewModel.LimitedViewModels) {
```
vmartos: LimitedViewModels is a collection of Playlists, but we need to get the element selection of each playlist and also we need to get the reference of the playlist to know the position of the selected elements to insert before or after that elements, finally a Save is needed on the Playlist.

VAS.Services/State/EditPlaylistElementState.cs
```diff
@@ -0,0 +1,39 @@
...
+
+		public override string Name {
+			get {
+				return NAME;
+			}
+		}
+
+		protected override void CreateViewModel (dynamic data)
+		{
+			ViewModel = data.PlaylistElement;
```
vmartos: It will crash, but it should crash since it should only work with PlaylistElements. The same happens with all states, if you pass incorrect data type it will crash.

VAS.Services/VideoPlayerController.cs
```diff
@@ -739,6 +735,7 @@ public virtual void LoadPlaylistEvent (Playlist playlist, IPlaylistElement eleme
 				LoadFrameDrawing (element as PlaylistDrawing, playing);
 			}
 			UpdateDuration ();
+			UpdatePlayingState (true);
```
vmartos: no, playing refers here on the playing state (paused or playing) but here we want to set the playlistElement.Playing to true even if is in paused state, because this changes the playlistElement.Playing to true and then we can render the Eye. The same behaviour is happening with the events, the Playing property is setted to true even if the playing state of the video is paused.

PR #332, opened 2017-07-06
Fixes VideoPlayer when working with events/playlists

vmartos:


VAS.Services/VideoPlayerController.cs
```diff
@@ -276,6 +276,7 @@ protected override void DisposeManagedResources ()
 				mediafileSet = value;
 				mediaFileSetCopy = value.Clone ();
 				if (mediafileSet != null) {
+					playerVM.FileSet = new MediaFileSetVM { Model = mediafileSet };
```
ylatuya: If FileSet is set here it should be remove from PlayerVM.Open

VAS.Services/VideoPlayerController.cs
```diff
@@ -1276,6 +1276,7 @@ protected virtual void InternalOpen (MediaFileSet fileSet, bool seek, bool force
 				ValidateVisibleCameras ();
 				UpdateZoom ();
 				UpdatePar ();
+				playerVM.ControlsSensitive = true;
```
ylatuya: Missing tests

PR #333, opened 2017-07-07
RA-994

SandraPG85:
Fix LimitedCVM notification not working properly

### DIFF
--- a/VAS.Core/MVVMC/LimitedCollectionViewModel.cs
+++ b/VAS.Core/MVVMC/LimitedCollectionViewModel.cs
@@ -37,6 +37,7 @@ namespace VAS.Core.MVVMC
 		where TModel : StorableBase
 	{
 		LicenseLimitationVM limitation;
+		bool notifying;
 
 		/// <summary>
 		/// Initializes a new instance of the <see cref="T:VAS.Core.MVVMC.LimitedCollectionViewModel`3"/> class.
@@ -46,6 +47,15 @@ namespace VAS.Core.MVVMC
 		{
 			this.SortByCreationDateDesc = sortByCreationDateDesc;
 			LimitedViewModels = new RangeObservableCollection<TViewModel> ();
+			LimitedViewModels.CollectionChanged += HandleViewModelsCollectionChanged;
+			base.ViewModels.CollectionChanged += HandleBaseViewModelModelsCollectionChanged;
+		}
+
+		protected override void DisposeManagedResources ()
+		{
+			base.DisposeManagedResources ();
+			LimitedViewModels.CollectionChanged -= HandleViewModelsCollectionChanged;
+			base.ViewModels.CollectionChanged += HandleBaseViewModelModelsCollectionChanged;
 		}
 
 		/// <summary>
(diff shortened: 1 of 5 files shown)


VAS.Core/MVVMC/LimitedCollectionViewModel.cs
```diff
@@ -46,6 +47,15 @@ public LimitedCollectionViewModel (bool sortByCreationDateDesc = true)
 		{
 			this.SortByCreationDateDesc = sortByCreationDateDesc;
 			LimitedViewModels = new RangeObservableCollection<TViewModel> ();
+			LimitedViewModels.CollectionChanged += HandleViewModelsCollectionChanged;
+			base.ViewModels.CollectionChanged += HandleModelsCollectionChanged;
```
ylatuya: The hanlders is ModelsCollectionChanged but you are attaching it to the ViewModels, also the base class provides a virtual handler that be overriden like you have done in the CollectionViewModel

VAS.Core/MVVMC/LimitedCollectionViewModel.cs
```diff
@@ -37,6 +37,7 @@ public class LimitedCollectionViewModel<TModel, TViewModel> : CollectionViewMode
 		where TModel : StorableBase
 	{
 		LicenseLimitationVM limitation;
+		bool notifying;
```
ylatuya: You have "editing" in the base class, can you use it maybe ?

SandraPG85: in this case I don't think so, because if use editing then the ModelsChange event will be skipped and the models will not be added.

VAS.Core/MVVMC/LimitedCollectionViewModel.cs
SandraPG85: I will update the handler name, but the subscription ha to be done to the base.ViewModels because if the collection is accessed trough the base class then the limited collection will not be updated. This could because an event of models change or accessing directly the view models of the base class.

VAS.Core/MVVMC/LimitedCollectionViewModel.cs
ylatuya: ok

VAS.Core/MVVMC/LimitedCollectionViewModel.cs
```diff
@@ -119,10 +129,39 @@ protected override void SetModel (RangeObservableCollection<TModel> model)
 			FillLimitedViewModels (base.ViewModels);
 		}
 
-		protected override void HandleViewModelsCollectionChanged (object sender, NotifyCollectionChangedEventArgs e)
+		protected new void HandleViewModelsCollectionChanged (object sender, NotifyCollectionChangedEventArgs e)
 		{
-			FillLimitedViewModels (base.ViewModels);
-			base.HandleViewModelsCollectionChanged (sender, e);
+			if (!notifying) {
```
ylatuya: As per the coding style we always prefer to bail out first to avoid extra indentation, in this case if (notifying) return;

PR #347, opened 2017-07-17
Add RetryIcon to styleconf and Red Color to show error

vmartos:


VAS.Core/StyleConf.cs
```diff
@@ -262,6 +263,10 @@ public class StyleConf
 
 		public const string MissingVideo = "icons/hicolor/scalable/actions/vas-missing-video" + IMAGE_EXT;
 
+		public const string RetryIcon = "vas-retry";
+
+		public static Color RedError = Color.Parse ("#AD474A");
```
ylatuya: Move to a property to StyleConf

PR #349, opened 2017-07-19
LON-1100: Load Welcome page with the icon of the proper longomatch edition

vmartos:


dfernandezramos:
The commit message does not match with the code has been done

dfernandezramos:
👍

PR #350, opened 2017-07-25
VAS-96

dfernandezramos:


dfernandezramos:
Still pending to do the same change in longomatch when this is approved

build/git-version-gen
```diff
@@ -2,242 +2,23 @@
...
-          case $v in
-            *-dirty) ;;
-            *) v="$v-dirty" ;;
-          esac ;;
-  esac
-fi
+# Read the version from the file
+while IFS='' read -r line || [[ -n "$line" ]]; do
+    v=`echo "$line"`
+done < "Version.txt"
```
vguzmanp: The version is always the first line, isn't it? It can be done with `IFS='' read -r line < "Version.txt"` directly.

build/git-version-gen
```diff
@@ -2,242 +2,23 @@
...
+
+# Recreate the number of commits and append it to the version from the file.
+vcommit=`git log -n 1 --pretty=format:%H -- Version.txt`
+commit_list=`git rev-list "$vcommit"..HEAD 2>/dev/null` \
+    || { commit_list=failed;
+         echo "WARNING: git rev-list failed" 1>&2; }
+numcommits=`echo "$commit_list" | wc -l`
+#1.1.1-3-gf76210a
+#1.1.1.3-f762
+v=`echo "$v"."$numcommits" | tr -d '[:space:]'`
```
vguzmanp: This doesn't add the `-<hash>` at the end, is that intentional?

build/git-version-gen
```diff
@@ -2,242 +2,23 @@
...
+done < "Version.txt"
+
+# Recreate the number of commits and append it to the version from the file.
+vcommit=`git log -n 1 --pretty=format:%H -- Version.txt`
+commit_list=`git rev-list "$vcommit"..HEAD 2>/dev/null` \
+    || { commit_list=failed;
+         echo "WARNING: git rev-list failed" 1>&2; }
+numcommits=`echo "$commit_list" | wc -l`
+#1.1.1-3-gf76210a
+#1.1.1.3-f762
```
vguzmanp: Why these two comments?

build/git-version-gen
```diff
@@ -6,19 +6,17 @@ scriptversion=2014-12-02.19; # UTC
 v=
 
 # Read the version from the file
-while IFS='' read -r line || [[ -n "$line" ]]; do
-    v=`echo "$line"`
-done < "Version.txt"
+IFS='' read -r line < "Version.txt"
+v=`echo "$line"`
```
vguzmanp: You don't need the `echo`. `line` is the variable name already. You can write `...read -r v <...`

vguzmanp: Or, to keep readability, `v=$line`

build/git-version-gen
```diff
@@ -2,242 +2,21 @@
...
-          esac ;;
-  esac
-fi
+# Recreate the number of commits and append it to the version from the file.
+vcommit=`git log -n 1 --pretty=format:%H -- Version.txt`
+commit_list=`git rev-list "$vcommit"..HEAD 2>/dev/null` \
+    || { commit_list=failed;
+         echo "WARNING: git rev-list failed" 1>&2; }
+numcommits=`echo "$commit_list" | wc -l`
+h=`git describe --abbrev=4 HEAD  | head -n1 | cut -d "-" -f3`
```
vguzmanp: Why use one-letter variables? `v` I understand because it was the one already in use, but `h`?

vguzmanp: `head` is useless here, as git describe returns a single line

build/git-version-gen
```diff
@@ -7,16 +7,16 @@ v=
...
 
 # Recreate the number of commits and append it to the version from the file.
 vcommit=`git log -n 1 --pretty=format:%H -- Version.txt`
 commit_list=`git rev-list "$vcommit"..HEAD 2>/dev/null` \
     || { commit_list=failed;
          echo "WARNING: git rev-list failed" 1>&2; }
 numcommits=`echo "$commit_list" | wc -l`
-h=`git describe --abbrev=4 HEAD  | head -n1 | cut -d "-" -f3`
-v=`echo "$v"."$numcommits"-"$h" | tr -d '[:space:]'`
+last_tag_hash=`git describe --abbrev=4 | head -n1 | cut -d "-" -f3`
```
vguzmanp: 😄 The `head` that I said is useless is the `head -n1`.

Funnily enough, the `HEAD` in the git command was also useless as it's implied if you don't pass it.

vguzmanp: what git describe is returning is not the last tag hash, it's the current commit's hash

PR #351, opened 2017-07-26
Ra 1031

vguzmanp:


VAS.UI.Gtk2/UI/Component/JobsManagerView.cs
```diff
@@ -105,6 +105,12 @@ public void OnLoad ()
 
 		public void OnUnload ()
 		{
+			ctx.Dispose ();
+			ctx = null;
+			ViewModel.Dispose ();
+			ViewModel = null;
+			treeview.Dispose ();
+			treeview = null;
```
vmartos: Maybe on the OnDestroyed or Dispose method?

VAS.Core/MVVMC/ControllerBase.cs
```diff
@@ -103,9 +104,42 @@ public class ControllerBase<TViewModel> : ControllerBase
 			}
 		}
 
+		public override async Task Start ()
+		{
+			await base.Start ();
+			if (ViewModel == null) {
+				throw new InvalidOperationException ($"The controller {GetType ()} needs a ViewModel before starting");
+			}
+			ConnectEvents ();
```
vguzmanp: It's just for safety. Forcing them to subscribe correctly means that we are sure that they are unsubscribed when needed.

If they don't want to subscribe they can just leave it there (so an empty handler will be subscribed) or override `ConnectEvents` and don't call the base method.

vguzmanp: All controllers that are tied to a ViewModel should be migrated to this structure, which ones are not doing it? I could migrate them all at the same time

VAS.Core/ViewModel/PlaylistCollectionVM.cs
```diff
@@ -38,12 +38,20 @@ public class PlaylistCollectionVM : LimitedCollectionViewModel<Playlist, Playlis
...
+
+		static void LoadIcons ()
+		{
 			deleteIcon = App.Current.ResourcesLocator.LoadIcon ("vas-delete", App.Current.Style.IconSmallWidth);
 			newIcon = App.Current.ResourcesLocator.LoadIcon ("vas-new-playlist", App.Current.Style.IconSmallWidth);
 		}
 
 		public PlaylistCollectionVM ()
 		{
+			if (deleteIcon == null || deleteIcon.Width == -1) {
```
vguzmanp: Yes... I'm not sure how to properly fix this actually.

The playlist subtoolbar has buttons with an icon that they get from here. When destroying the button, the image is destroyed as well, regardless of it being static. Then the next time you use the icon, it calls the already disposed icon, and crashes.

I tried removing the image from the buttons before the dispose, but it didn't work.

VAS.Core/ViewModel/MediaFileSetVM.cs
```diff
@@ -28,22 +28,32 @@ namespace VAS.Core.ViewModel
 	/// </summary>
 	public class MediaFileSetVM : CollectionViewModel<MediaFile, MediaFileVM>, IViewModel<MediaFileSet>
 	{
-		MediaFileSet model;
+		new MediaFileSet model;
```
vguzmanp: Done

VAS.Core/MVVMC/ControllerBase.cs
vmartos: AnalysisController is tied to a VM but subscribes only to the FileSet and Timeline VM, you can override here the ConnectEvents.

In fact most of the Controllers just inherit from ControllerBase even if it receives a VM, Then fill up all of it's ViewModels and then they subscribe to property changes. So i wonder if we can move the ConnectEvents to the ControllerBase (being empty), then ControllerBase<VM> override to connect to its VM. And so later all of Controllers that inherits from ControllerBase could override the ConnectEvents to connect all of his VM's to the property changes, but that is only an idea, because now it will require quite a few changes to all controllers.

For example TaggingController when setting the VideoPlayerVM it subscribes to Property changes but then it do not unsubscribe in Stop, if we make the ConnectEvents available to all controllers we can just apply this structure to all the controllers

VAS.Core/ViewModel/PlaylistCollectionVM.cs
vmartos: ok, for me is ok then

PR #352, opened 2017-07-28
LON-1039

vmartos:


VAS.Core/License/LicenseLimitation.cs
```diff
@@ -20,45 +20,15 @@
...
-	/// </summary>
 	public class LicenseLimitation : BindableBase
 	{
-		int maximum;
-
 		/// <summary>
 		/// Gets or sets the name of the limitation.
 		/// </summary>
 		/// <value>The name of the limitation.</value>
 		public string Name { get; set; }
```
SandraPG85: Add a FeatureLimitation where the feature description can be set to display in the upgrade dialog.
Also rename the Name property in the LicenseLimitation for a name that refers the registration purpose of the field.
Update the FeatureLimitationVM in consequence.

VAS.Services/LicenseLimitationsService.cs
```diff
@@ -116,17 +130,74 @@ public void Add (LicenseLimitation limitation, Command command = null)
...
+		/// Checks if a limitation feature can be executed
+		/// </summary>
+		/// <param name="name">Name of the feature limitation</param>
+		public bool CanExecuteFeature (string name)
+		{
+			var featureLimitVM = Get<FeatureLimitationVM> (name);
+
+			if (featureLimitVM == null) {
+				throw new InvalidOperationException ("Cannot get Feature, because it wasn't registered," +
+													 "register the feature prior calling this method");
```
SandraPG85: As we discussed in the daily meeting, if the feature has not been registered then it is not limited and it can be executed.

VAS.Services/ViewModel/UpgradeLimitationVM.cs
```diff
@@ -0,0 +1,62 @@
...
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using System;
+using VAS.Core.Common;
+using VAS.Core.MVVMC;
+using VAS.Core.ViewModel;
+
+namespace VAS.Services.ViewModel
+{
+	public class UpgradeLimitationVM : ViewModelBase
```
SandraPG85: Comment the class and the public properties

VAS.UI.Gtk2/UI/Panel/UpgradeLimitationPanel.cs
```diff
@@ -0,0 +1,213 @@
...
+			get {
+				return viewModel;
+			}
+
+			set {
+				viewModel = value;
+				if (viewModel != null) {
+					AddFeatures ();
+					ConfigureSeeOtherPlans ();
+					ctx.UpdateViewModel (viewModel);
```
SandraPG85: The add features and configureotherplans can be moved to the load function?

SandraPG85: If the viewnodel changes, the old features are removed?

vmartos: If i move to Load function it doesn't render the features, don't tell me why -_-... but when a new ViewModel is set it removes the old features before setting the new ones.

PR #354, opened 2017-08-03
Lon 1042

SandraPG85:
Add excel export limitation


VAS.Core/Addins/ProjectExporterBase.cs
```diff
@@ -45,7 +45,13 @@ public abstract class ProjectExporterBase : IProjectExporter
 			get;
 		}
 
-		protected abstract void ExportProject (Project project, string filename);
+		/// <summary>
+		/// Gets or sets a value indicating whether the exporta
```
vmartos: Fix Documentation

PR #357, opened 2017-08-07
Lon 1035

vguzmanp:


VAS.Core/License/FeatureLicenseLimitation.cs
```diff
@@ -24,9 +24,5 @@ namespace VAS.Core.License
 	/// </summary>
 	public class FeatureLicenseLimitation : LicenseLimitation
 	{
-		public string FeatureName {
-			get;
-			set;
-		}
 	}
```
SandraPG85: Why to remove it?

VAS.Core/License/LicenseLimitation.cs
```diff
@@ -26,13 +26,30 @@ namespace VAS.Core.License
...
 		/// <summary>
+		/// Gets or sets the display name of the limitation.
+		/// This is the one that will be displayed in popups, etc.
+		/// If this is not set, it returns the RegisterName.
+		/// </summary>
+		/// <value>The display name of the limitation.</value>
+		public string DisplayName {
+			get {
+				return displayName ?? RegisterName;
+			}
```
SandraPG85: The feature name should be translated, so I will not use as default value the one used to register the limitation

VAS.Core/ViewModel/LimitationVM.cs
```diff
@@ -72,9 +82,9 @@ public class FeatureLimitationVM : LimitationVM
 		/// Gets the name of the feature.
 		/// </summary>
 		/// <value>The name of the feature.</value>
-		public string FeatureName {
+		public string DisplayName {
 			get {
```
SandraPG85: if the LimitationVM already has DisplayName this one can be removed

VAS.Drawing/CanvasObjects/Blackboard/ImageCanvasObject.cs
```diff
@@ -0,0 +1,38 @@
...
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using System;
+using VAS.Core.Common;
+using VAS.Core.Interfaces.Drawing;
+using VAS.Drawing.CanvasObjects;
+
+namespace VAS.Drawing.CanvasObjects.Blackboard
+{
+	public class ImageCanvasObject : CanvasObject
```
SandraPG85: add class and properties documentation

VAS.Services/LicenseLimitationsService.cs
```diff
@@ -149,19 +149,25 @@ public void Add (FeatureLicenseLimitation limitation)
...
-													 " returning true");
+			var limitationVM = Get<LimitationVM> (name);
+			if (limitationVM == null) {
+				Log.Warning ("Cannot get limitation because it wasn't registered." +
+													 " Returning true");
 				return true;
 			}
-			return !featureLimitVM.Enabled;
+			if (!limitationVM.Enabled) {
+				return true;
```
SandraPG85: this check can be removed since at the end the return will do the same thing

VAS.Core/License/FeatureLicenseLimitation.cs
vguzmanp: It's replaced with DisplayName now

VAS.Core/License/LicenseLimitation.cs
vguzmanp: hmm I'm not sure if I follow you here. This is a fallback in case DisplayName is not defined. When defined, DisplayName should always be translated

VAS.Core/ViewModel/LimitationVM.cs
vguzmanp: I don't follow you here, this is the LimitationVM.

vguzmanp: nevermind

VAS.Services/LicenseLimitationsService.cs
vguzmanp: Not really, if it's a `CountLimitationVM` and it's not enabled, I want it to return before. Otherwise it would return with `countVM.Count < countVM.Maximum`

VAS.Core/MVVMC/BindingsExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace VAS.Core.MVVMC
+{
+	public static class BindingsExtensions
+	{
+		/// <summary>
+		/// Bind the specified label to a property by name.
+		/// </summary>
+		/// <param name="label">Label.</param>
+		public static PropertyToPropertyBinding<TProperty> Bind<TProperty> (this IViewModel label, Expression<Func<IViewModel, TProperty>> setterExpression, Expression<Func<IViewModel, TProperty>> propertyExpression)
+		{
```
SandraPG85: why the IViewModel argument is called label?

PR #358, opened 2017-08-08
Unbind the buttonBinding correctly

vguzmanp:


VAS.UI.Helpers.Gtk2/Bindings/ToggleButtonBinding.cs
```diff
@@ -57,6 +57,14 @@ protected override void BindViewModel ()
 			}
```
aameijide: Use unbindViewModel to unsubscribe CanExecuteChanged

VAS.Tests/MVVMC/TestBinding.cs
```diff
@@ -90,5 +91,81 @@ public void TestBindNewViewModel ()
...
+			var binding = button.Bind ((vm) => new Command ((obj) => executed = true));
+			binding.ViewModel = new ViewModelBase ();
+
+			// Act
+			button.Click ();
+
+			// Assert
+			Assert.IsTrue (executed);
+		}
+
```
aameijide: Check also that if a command that cannot execute is clicked, this is not executed

PR #359, opened 2017-08-08
LON-1048

SandraPG85:
Add linking buttons limitation

### DIFF
--- a/VAS.Tests/MVVMC/TestLimitationCommand.cs
+++ b/VAS.Tests/MVVMC/TestLimitationCommand.cs
@@ -138,6 +138,7 @@ namespace VAS.Tests.MVVMC
 
 			// Assert
 			Assert.IsTrue (executed);
+			Assert.AreEqual (executed, command.Executed);
 		}
 
 		[Test]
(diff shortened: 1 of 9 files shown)


VAS.Tests/MVVMC/TestLimitationCommand.cs
```diff
@@ -138,6 +138,7 @@ public void Execute_LimitationNotInitialized_Executed ()
 
 			// Assert
 			Assert.IsTrue (executed);
+			Assert.AreEqual (executed, command.Executed);
```
aameijide: I don't get why we need this assert if we have Assert.IsTrue(executed), is not the same?

VAS.Tests/MVVMC/TestLimitationCommand.cs
```diff
@@ -169,6 +170,7 @@ public void Execute_LimitationCanNotExecute_MoveToUpgradeDialog ()
 
 			// assert
 			Assert.IsFalse (executed);
+			Assert.AreEqual (executed, command.Executed);
```
aameijide: Same as above: I don't get why we need this assert if we have Assert.IsTrue(executed), is not the same?

VAS.Tests/MVVMC/TestLimitationCommand.cs
SandraPG85: we are ensuring that the command executed has been set by the limitation command, without this check we don't know

PR #360, opened 2017-08-10
Vas 110: Migration to MONO5

vmartos:


build/build.mk
```diff
@@ -5,7 +5,7 @@ OUTPUT_DIR ?= $(libdir)/@USER_PACKAGE@
 XBUILD_CMD = $(XBUILD) \
  /property:DefineConstants="$(XBUILD_CONSTANTS)" \
  /property:BuildType=makefiles \
- /property:BuildProjectReferences=false \
+ /property:BuildProjectReferences=true \
```
vguzmanp: Why is this?

vmartos: there was a problem when trying to build with msbuild and property:BuildProjectReferences seems that there wasn't building the projects and was finding for e.g. VAS.Core.dll instead of building it. with xbuild I think it should work with false, so i will remove this change

PR #363, opened 2017-08-21
Lon 1133

vguzmanp:


VAS.Services/Controller/ProjectsController.cs
```diff
@@ -183,23 +183,24 @@ Task HandleNew (CreateEvent<TModel> evt)
 			// >> Cloning a preloaded a project
 			// >> Using the stored viewmodels instead of creating new ones
 
-			/*if (projectVM != null) {
+			if (projectVM != null) {
 				if (ViewModel.LoadedProject.Edited == true) {
```
SandraPG85: Enabling this again will introduce a poor user experience when changing between projects.
The clone could be removed doing the view model stateful, so the projects changes are keep in the view model properties and only stored under demand.

VAS.Core/ViewModel/CountLimitationBarChartVM.cs
```diff
@@ -15,6 +15,8 @@
 //  along with this program; if not, write to the Free Software
 //  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
```
SandraPG85: add test for the vm

VAS.Services/Controller/MediaFileSetController.cs
```diff
@@ -0,0 +1,83 @@
...
+using VAS.Core.Events;
+using VAS.Core.Interfaces.MVVMC;
+using VAS.Core.MVVMC;
+using VAS.Core.Store;
+using VAS.Core.ViewModel;
+
+namespace VAS.Services.Controller
+{
+	public class MediaFileSetController : ControllerBase
+	{
```
SandraPG85: comment class

PR #367, opened 2017-08-24
Order by start time when loading a list of events

vmartos:
-Also add tests


VAS.Tests/Services/TestEventsController.cs
```diff
@@ -149,6 +193,21 @@ public void TestLoadEventTypeTimelineVM ()
 		}
 
 		[Test]
+		public void LoadEventTypeTimelineVM_OrderedByStartTime ()
```
dfernandezramos: PlayerController_LoadEventTypeTimelineVM_OrderedByStartTime

PR #368, opened 2017-08-28
Lon 894

vmartos:


VAS.Drawing/Widgets/DashboardCanvas.cs
```diff
@@ -443,6 +427,35 @@ void AddButton (DashboardButtonVM vm)
 			vm.ActionLinks.ViewModels.CollectionChanged += HandleActionLinksCollectionChanged;
 		}
 
+		void AddActionLinks (DashboardButtonView buttonObject) {
```
vguzmanp: This should go to a controller

VAS.Services/Controller/DashboardEditorController.cs
```diff
@@ -59,15 +66,97 @@ public class DashboardEditorController : ControllerBase
...
+			//Remove Dest ActionLinks
+			foreach (var b in ViewModel.ViewModels) {
+				var linksToRemove = b.ActionLinks.Where (al => al.DestinationButton == button);
+				if (linksToRemove.Any ()) {
+					b.ActionLinks.ViewModels.RemoveRange (linksToRemove);
+				}
+			}
+		}
+
+		void RemoveActionLinks (Tag tag)
```
vguzmanp: Why not use a VM here?

VAS.Services/Controller/DashboardEditorController.cs
```diff
@@ -59,15 +66,97 @@ public class DashboardEditorController : ControllerBase
...
+			foreach (var b in ViewModel.ViewModels) {
+				var linksToRemove = b.ActionLinks.Where (al => SameReferenceTags(al.SourceTags, tag) || SameReferenceTags(al.DestinationTags, tag));
+				if (linksToRemove.Any ()) {
+					b.ActionLinks.ViewModels.RemoveRange (linksToRemove);
+				}
+			}
+		}
+
+		/// <summary>
+		/// Sames the reference tags. We need a way to compare Tags by same reference, since <see cref="Tag"/> object
```
vguzmanp: `Sames the reference tags`? 🤔

vguzmanp: The method is more like `Contains` isn't it?

VAS.Services/Controller/DashboardEditorController.cs
```diff
@@ -33,9 +35,14 @@ namespace VAS.Services.Controller
 	/// <summary>
 	/// A controller to edit a <see cref="Dashboard"/>.
 	/// </summary>
-	public class DashboardEditorController : ControllerBase
+	public class DashboardEditorController : ControllerBase<DashboardVM>
 	{
-		DashboardVM dashboardVM;
+		List<AnalysisEventType> eventTypes;
```
vguzmanp: Controllers should be stateless, and get the state from the viewModel. Why do we need to store this list here?

VAS.Services/Controller/DashboardEditorController.cs
```diff
@@ -153,16 +250,24 @@ void ReplaceField (FieldPositionType fieldPosition, Image background = null)
...
 			case FieldPositionType.Goal:
-				dashboardVM.GoalBackground = background ?? App.Current.GoalBackground;
+				ViewModel.GoalBackground = background ?? App.Current.GoalBackground;
 				break;
 			}
 		}
 
+		void HandleTagsCollectionChanged (object sender, NotifyCollectionChangedEventArgs e)
+		{
+			if (e.Action == NotifyCollectionChangedAction.Remove) {
```
vguzmanp: What happens in the other cases?

VAS.Services/Controller/DashboardEditorController.cs
vmartos: This is Stateless, it is just a helper list to store all eventTypes of the actual selected dashboard. I subscribe to PropertyChanges in the ViewModel (Dashboard) and when it changes the model i need to register to ObservableCollection Tags changes from all eventTypes in order to remove the ActionLinks from the related buttons.

See that that list is rebuilded every time the Model changes.

VAS.Services/Controller/DashboardEditorController.cs
```diff
@@ -59,15 +66,97 @@ public class DashboardEditorController : ControllerBase
 			await base.Stop ();
 		}
 
+		protected override void ConnectEvents ()
+		{
+			base.ConnectEvents ();
+			ConnectEventTypes ();
```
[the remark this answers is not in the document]
vmartos: because when the model is changed we need to connect to the new EventTypes, so i refactor the connection to the eventypes in a method to be called from 2 different places.

VAS.Services/Controller/DashboardEditorController.cs
vmartos: a view model for the Tag doesn't exist yet.

VAS.Services/Controller/DashboardEditorController.cs
vmartos: yep, but we need to check that is actually the same reference in order to know that we have to remove the action links related to that tag. Because the Tag class overrides the Equals method and compares with Value and Group, but we can have different tags with the same Value and group ex: Group:Outcome, Value: Success

VAS.Services/Controller/DashboardEditorController.cs
vmartos: Here we had the bug that when we remove a tag from an EventType the ActionLinks related to that tag where not removed, with this we "observe" the tags changes in order to remove the actionLinks. That was the purpose of the Fix. Since the Dialog that handles that logic is not migrated to MVVMC and that is working with EventTypes directly instead of DashboardButtons the only possible solution is to observe the Tag changes.

VAS.Services/Controller/DashboardEditorController.cs
vguzmanp: 👌

VAS.Services/Controller/DashboardEditorController.cs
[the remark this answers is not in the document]
vguzmanp: 👌

VAS.Services/Controller/DashboardEditorController.cs
vguzmanp: Could you create it? The more we use VM, the less things left To Do. And this VM shouldn't be difficult to create.

VAS.Services/Controller/DashboardEditorController.cs
vguzmanp: 👌 Add a FIXME that makes clear that this should be removed after the other dialog is ported to MVVM

VAS.Core/ViewModel/ActionLinkVM.cs
```diff
@@ -30,9 +30,15 @@ public class ActionLinkVM : ViewModelBase<ActionLink>
...
-		{
-			var k = 0;
+		public override ActionLink Model {
+			get {
+				return base.Model;
+			}
+			set {
+				base.Model = value;
+				SourceTags.Model = value.SourceTags;
+				DestinationTags.Model = value.DestinationTags;
```
vguzmanp: Use `value?.`

VAS.Core/ViewModel/ActionLinkVM.cs
```diff
@@ -85,10 +90,9 @@ public ActionLinkVM ()
 		/// <summary>
 		/// A list of tags that needs to be set in the destination
 		/// </summary>
-		public ObservableCollection<Tag> DestinationTags {
-			get { return Model.DestinationTags; }
-			set { Model.DestinationTags = value; }
-		}
+		public CollectionViewModel<Tag, TagVM> DestinationTags {
+			get;
+		} = new CollectionViewModel<Tag, TagVM> ();
```
vguzmanp: Do this in the constructor.

VAS.Core/ViewModel/ActionLinkVM.cs
```diff
@@ -113,5 +117,10 @@ public ActionLinkVM ()
 		public string Name {
 			get { return string.Format ("{0} -> {1}", SourceButton?.Name, DestinationButton?.Name); }
 		}
+
+		protected override void ForwardPropertyChanged (object sender, System.ComponentModel.PropertyChangedEventArgs e)
+		{
+			base.ForwardPropertyChanged (sender, e);
```
vguzmanp: Useless override

VAS.Core/ViewModel/DashboardButtonVM.cs
```diff
@@ -245,17 +247,32 @@ public DashboardButtonVM ()
...
 		Time currentTime;
+		TimedDashboardButton realModel;
+
+		/// <summary>
+		/// Gets the correctly Typed Model
+		/// </summary>
+		/// <value>The button.</value>
+		public TimedDashboardButton TypedModel {
+			get {
+				return realModel;
```
vguzmanp: Why not just return `(TimedDashboardButton)base.Model`?

VAS.Core/ViewModel/DashboardButtonVM.cs
```diff
@@ -355,41 +388,52 @@ public class EventButtonVM : TimedDashboardButtonVM
 		public EventTypeVM EventType {
 			get;
 			private set;
-		}
+		} = new EventTypeVM ();
```
vguzmanp: Do this in the constructor

VAS.Core/ViewModel/DashboardButtonVM.cs
```diff
@@ -420,16 +464,23 @@ public AnalysisEventButtonVM ()
...
-			set;
+		public RangeObservableCollection<TagVM> SelectedTags {
+			get {
+				return Tags.Selection;
+			}
+		}
+
+		public Dictionary<string, List<TagVM>> TagsByGroup {
+			get {
+				return Tags?.GroupBy (t => t.Group).ToDictionary (g => g.Key, g => g.ToList ());
```
vguzmanp: Is `Tags` ever null?

vguzmanp: Also, what is this for?

VAS.Drawing/CanvasObjects/Dashboard/ActionLinkView.cs
```diff
@@ -139,10 +139,10 @@ public bool CanLink (LinkAnchorView dest)
 
 			// FIXME: View using Model
```
vguzmanp: It's no longer FIXME, right? 😄

VAS.Drawing/CanvasObjects/Dashboard/AnalysisEventButtonView.cs
```diff
@@ -545,7 +545,7 @@ void DrawSelectedTags (IDrawingToolkit tk)
 			}
 			foreach (Rectangle r in rects.Keys) {
 				object obj = rects [r];
-				if (obj is Tag && SelectedTags.Contains (obj as Tag)) {
+				if (obj is Tag && SelectedTags.Contains (obj as TagVM)) {
```
vguzmanp: `obj is Tag && obj as TagVM`

VAS.Services/Controller/DashboardEditorController.cs
```diff
@@ -37,11 +37,11 @@ namespace VAS.Services.Controller
 	/// </summary>
 	public class DashboardEditorController : ControllerBase<DashboardVM>
 	{
-		List<AnalysisEventType> eventTypes;
+		List<AnalysisEventButtonVM> eventButtons;
```
vguzmanp: `EventType -> EventButtonVM`, is that correct?

vmartos: yes, but it was AnalysisEventType that is the Model that contains the list of Tags, so we need the AnalysisEventTypeButtonVM

VAS.Core/ViewModel/DashboardButtonVM.cs
vmartos: Q1: Never null, changed. 
Q2: It is used in AnalysisEventButtonView but just for grouping to help logic. They where using the AnalysisEventType.TagsByGroup instead. but i changed to use the TagVM's

PR #376, opened 2017-08-31
Limit the Zoom in the DrawingTool and BlackBoard

vmartos:


VAS.Drawing/Widgets/Blackboard.cs
```diff
@@ -450,7 +461,8 @@ protected override void StartMove (Selection sel)
 					}
 					newZoom = Math.Max (newZoom, MinZoom);
 					newZoom = Math.Min (newZoom, MaxZoom);
-					Zoom (newZoom, MoveStart);
+					centerZoom = MoveStart;
+					ZoomCommand.Execute (newZoom);
```
vguzmanp: Can't you send the center to the command, like it was before?

VAS.Services/ViewModel/DrawingToolVM.cs
```diff
@@ -75,11 +88,24 @@ public class DrawingToolVM : ViewModelBase
 		}
 
 		/// <summary>
+		/// Gets or sets the zoom level.
+		/// </summary>
+		/// <value>The zoom level.</value>
+		public double ZoomLevel {
```
vguzmanp: Where was this stored before?

VAS.UI.Gtk2/UI/Dialog/DrawingTool.cs
```diff
@@ -217,13 +219,18 @@ protected override void OnDestroyed ()
...
+				}
 				viewModel = value;
 				if (viewModel != null) {
+					viewModel.PropertyChanged += HandleViewModelPropertyChanged;
 					if (viewModel.TimelineEvent != null) {
 						LoadPlay (viewModel.TimelineEvent, viewModel.Frame, viewModel.Drawing, viewModel.CameraConfig);
 					} else {
 						LoadFrame (viewModel.Frame);
 					}
+					ctx.UpdateViewModel (viewModel);
```
vguzmanp: The ctx should be updated even when viewModel == null. That way the bindings are disconnected.

VAS.UI.Gtk2/UI/Dialog/DrawingTool.cs
```diff
@@ -389,7 +396,10 @@ public void SetViewModel (object viewModel)
 			blackboard.Background = frame;
 			savetoprojectbutton.Visible = true;
 			blackboard.Drawing = drawing;
-			blackboard.RegionOfInterest = drawing.RegionOfInterest;
+			//FIXME: this needs to show a warning message?
```
vguzmanp: Create a Task / Bug / Whatever so that this isn't forgotten before the release

VAS.UI.Gtk2/UI/Dialog/DrawingTool.cs
```diff
@@ -405,6 +415,12 @@ public void LoadFrame (Image frame)
 			UpdateTextSize ();
 		}
 
+		void Bind()
+		{
+			ctx = new BindingContext ();
+			ctx.Add (zoomscale.Bind (vm => ((DrawingToolVM)vm).SetZoomCommand, 1, 1, 4, 0.2, 0.2));
```
vguzmanp: What are these magic numbers?

VAS.UI.Gtk2/UI/Dialog/DrawingTool.cs
```diff
@@ -799,5 +805,15 @@ DrawTool GetTool (RadioButton sender)
 			}
 			return tool;
 		}
+
+		void HandleViewModelPropertyChanged (object sender, System.ComponentModel.PropertyChangedEventArgs e)
```
vguzmanp: using

VAS.UI.Helpers.Gtk2/Bindings/BindingsExtensions.cs
```diff
@@ -190,5 +190,65 @@ public static ComboBoxBinding Bind (this ComboBox comboBox, Expression<Func<IVie
...
+		}
+
+		/// <summary>
+		/// Bind the specified hscale, commandFunc and defaultValue. For LimitationCommand<double>
+		/// </summary>
+		/// <returns>The bind.</returns>
+		/// <param name="hscale">Hscale.</param>
+		/// <param name="commandFunc">Command func.</param>
+		/// <param name="defaultValue">Default value.</param>
+		public static HScaleCommandBinding Bind (this HScale hscale, Func<IViewModel, LimitationCommand<double>> commandFunc,
```
vguzmanp: Shouldn't the `Command` one be enough?

VAS.UI.Helpers.Gtk2/Bindings/HScaleCommandBinding.cs
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		protected override void UnbindView ()
+		{
+			hScale.ValueChanged -= HandleValueChanged;
+		}
+
+		protected override void BindViewModel ()
+		{
+			UnbindViewModel ();
```
vguzmanp: Shouldn't this be in the base class?

VAS.UI.Helpers.Gtk2/Bindings/HScaleCommandBinding.cs
```diff
@@ -0,0 +1,104 @@
...
+		{
+			if(skipCommand) {
+				skipCommand = false;
+				return;
+			}
+			Command.Execute (hScale.Value);
+
+			var limitedCommand = Command as LimitationCommand;
+			if (limitedCommand != null && !limitedCommand.Executed) {
+				skipCommand = true;
```
vguzmanp: Why is this?

VAS.UI.Gtk2/UI/Dialog/DrawingTool.cs
vmartos: done

VAS.UI.Helpers.Gtk2/Bindings/BindingsExtensions.cs
vmartos: the idea here is that an HScale has double values, so the binding should be Command<double> instead of Command, and LimitationCommand<T> doesn't inherit from Command<T>, it inherits from LimitationCommand. this was done to share the logic from the LimitationCommand.

VAS.Drawing/Widgets/Blackboard.cs
vmartos: the Command.Execute only accepts one parameter :(. But i can create a struct called Zoom (or something like that) that contains a double zoom value and Point center.

VAS.Services/ViewModel/DrawingToolVM.cs
vmartos: this was stored in the zoomscale.Value, and in the ValueChanged event the zoomLabel text was updated, now it is updated based on the ZoomLevel property in the ViewModel that the HScale binded with the command updates the ZoomLevel property.

VAS.UI.Helpers.Gtk2/Bindings/HScaleCommandBinding.cs
vmartos: because when the command could not be executed (invalid license) we should restore the HScale to the default position. In the example of zoom we change the zoom using the HScale then it visually changes to the middle of the scale, but the zoom couldn't be executed then it was maintaned in the middle but there was no zoom. With visually restore it to the default value and without trying to execute the command again.

VAS.Drawing/Widgets/Blackboard.cs
vguzmanp: Hmm, I don't really like creating a container for that. We should have a way to pass a list of arguments, and let the delegate choose what those are.

Leave it like this meanwhile.

VAS.UI.Gtk2/UI/Dialog/DrawingTool.cs
```diff
@@ -45,6 +46,10 @@ public abstract partial class DrawingTool : Gtk.Dialog, IPanel<DrawingToolVM>
 	{
 		const int MOVE_OFFSET = 5;
 		const int TOOL_HEIGHT = 24;
+		const double MIN_ZOOM = 1;
+		const double MAX_ZOOM = 4;
```
vguzmanp: Is x4 the max? I think the upgrade text said x5

VAS.UI.Helpers.Gtk2/Bindings/BindingsExtensions.cs
vguzmanp: Uff... I hate not having multiple inheritance to allow these things :(

VAS.Drawing/Widgets/Blackboard.cs
vmartos: ok! i'll add a fixme here to remember that when the commands accepts more than one argument it should add the MoveStart to the arguments and remove the centerZoom

VAS.UI.Helpers.Gtk2/Bindings/HScaleCommandBinding.cs
vguzmanp: Can't you call `UnbindView`, modify it, and rebind the view?

VAS.Drawing/Widgets/Blackboard.cs
vguzmanp: perfect 👌

VAS.UI.Helpers.Gtk2/Bindings/HScaleCommandBinding.cs
vmartos: I have tried and still enters the event handler, don't know why maybe you can't detach from an event handler inside the event handler method.

VAS.UI.Gtk2/UI/Dialog/DrawingTool.cs
vmartos: maybe the text is incorrect, This was previously at 4 before my changes, and in the video seems that is 4x also (even in the zoom at the video says 200% :S)

PR #384, opened 2017-09-07
Ra 1057

vguzmanp:


VAS.DB/CouchbaseManager.cs
```diff
@@ -154,7 +154,7 @@ protected IStorage Add (string name, bool check)
 				return db;
 			} catch (CouchbaseLiteException ex) {
 				Log.Exception (ex);
```
ylatuya: Maybe we don't need to try catch here anymore.

PR #389, opened 2017-09-18
Lon 1037

vmartos:


VAS.Core/MVVMC/LimitedCollectionViewModel.cs
```diff
@@ -173,21 +174,22 @@ protected void HandleBaseViewModelModelsCollectionChanged (object sender, Notify
...
 				return;
 			}
-			notifying = true;
 			FillLimitedViewModels (base.ViewModels);
-			notifying = false;
 		}
 
-		void FillLimitedViewModels (IEnumerable<TViewModel> viewmodels)
+		protected virtual void FillLimitedViewModels (IEnumerable<TViewModel> viewmodels)
 		{
```
SandraPG85: Instead of overriding this method create a new one to specify static objects that cannot be limited

VAS.Core/ViewModel/CountLimitationBarChartVM.cs
```diff
@@ -103,21 +124,39 @@ public void Bind ()
...
 
+		void UpdateVisibility ()
+		{
+			if (Limitation.Remaining <= showOnRemaining && showOnRemaining != -1) {
+				Visible = Limitation.Enabled;
+			} else if (showOnRemaining != -1) {
+				Visible = false;
+			} else {
+				Visible = Limitation.Enabled;
+			}
```
SandraPG85: this code could be simplified assigning at the beginning first the Visible = Limitation.Enabled and after the check to set it to false.

VAS.Core/ViewModel/TimelineVM.cs
```diff
@@ -347,6 +391,33 @@ void AddToPlayersTimeline (TimelineEventVM timelineEventVM)
...
+					} else {
+						playerToEvent [player].Add (timeline);
+					}
+				}
+			}
+
+			foreach (var player in playerToEvent) {
+				var timelineEvents = playerToTimeline [player.Key].ViewModels;
+				timelineEvents.RemoveRange (timelineEvents.Except (player.Value));
+				timelineEvents.AddRange (player.Value.Except (timelineEvents));
```
SandraPG85: RangeObservableCollection will add a Reset function and a Replace function where the collection is not being recreated, you can use it in here.

VAS.Core/ViewModel/TimelineVM.cs
```diff
@@ -43,6 +43,7 @@ public class TimelineVM : ViewModelBase<RangeObservableCollection<TimelineEvent>
 		RangeObservableCollection<TimelineEvent> model;
 		Dictionary<string, EventTypeTimelineVM> eventTypeToTimeline;
 		Dictionary<Player, PlayerTimelineVM> playerToTimeline;
+		CountLimitationBarChartVM limitationChart;
```
SandraPG85: Are these changes validated in some test, if not add a test for the timelineVM

VAS.Core/MVVMC/LimitedCollectionViewModel.cs
SandraPG85: I have modified the logic in base class, to be able to integrate the changes a new function must be created to configure which view models must be filtered, since this will happen also in the dash boar

VAS.Core/ViewModel/TimelineVM.cs
vmartos: done

VAS.Core/MVVMC/LimitedCollectionViewModel.cs
vmartos: done

VAS.Core/MVVMC/LimitedCollectionViewModel.cs
```diff
@@ -185,7 +197,8 @@ void FilterLimitedViewModels (IEnumerable<TViewModel> viewmodels, bool sorted =
 			bool filterApplied = false;
 			if (Limitation?.Enabled ?? false) {
 				int previousViewmodels = viewmodels.Count ();
-				viewmodels = viewmodels.Take (Limitation.Maximum);
+				viewmodels = viewmodels.Except (StaticViewModels).Take (Limitation.Maximum);
+				viewmodels = viewmodels.Union (StaticViewModels);
 				filterApplied = viewmodels.Count () != previousViewmodels;
```
SandraPG85: Add test to validate this static view models change

vmartos: done

PR #398, opened 2017-09-22
Move the TimeLineEvent compare method to its VM

dfernandezramos:


VAS.Core/ViewModel/TimelineEventVM.cs
```diff
@@ -118,6 +118,48 @@ public class TimelineEventVM : TimeNodeVM, IViewModel<TimelineEvent>
 				Model.FieldPosition = value;
 			}
 		}
+
+		/// <summary>
+		/// Compare the specified timelineEventVM with the calling one.
```
SandraPG85: this vm should implement IComparable interface

PR #401, opened 2017-09-26
Lon 1185

ylatuya:


VAS.Tests/Services/TestDrawingsController.cs
```diff
@@ -0,0 +1,248 @@
...
+		}
+
+		[TearDown]
+		public async Task TearDown ()
+		{
+			await controller.Stop ();
+		}
+
+		[Test]
+		public void SetViewModel_VideoPlayerOnly ()
```
dfernandezramos: Remember that the tests names must be ThingTested_ActionPerformed_ExpectedResult. You have forgot the expected result. Here, for example, would be something like SetViewModelNotThrown

PR #402, opened 2017-09-27
LON-1197

dfernandezramos:


VAS.UI.Multimedia.Gtk2/UI/VideoPlayerView.cs
```diff
@@ -751,7 +751,28 @@ void HandleJumpsButtonClicked (object sender, System.EventArgs e)
 		void HandleReady (object sender, EventArgs e)
 		{
 			playerVM.ViewPorts = new List<IViewPort> { mainviewport, subviewport1.Viewport, subviewport2.Viewport, subviewport3.Viewport };
-			playerVM.Ready (true);
+			switch (playerVM.ViewPorts.Where (vp => vp.WindowHandle != null).Count ()) {
```
ylatuya: You should use the MediaFileSet to list the number of view ports that should be ready.

if (!playerVM.ViewPorts.Take (playerVM.FileSet.Count).Any (p => !p.Ready)) {
    playerVM.Ready (true);
}

ylatuya: And you probably don't have to change the interface this way

dfernandezramos: The interface must be changed due because we need to lack the get accesor for the ViewPorts property `playerVM.ViewPorts.Take`.

PR #403, opened 2017-09-27
Add a MessageToDisplay to the licenseStatus

vmartos:
- this is useful to display a message after the check has done


VAS.Core/Interfaces/License/ILicenseStatus.cs
```diff
@@ -58,5 +58,14 @@ public interface ILicenseStatus
 		IEnumerable<string> Limitations {
 			get;
 		}
+
+		/// <summary>
+		/// Gets or sets the message to display.
+		/// </summary>
+		/// <value>The message to display.</value>
+		string MessageToDisplay {
```
ylatuya: Why not simply Message?

PR #404, opened 2017-09-29
LON-1199

dfernandezramos:
Create a command to limit saving a static template

### DIFF
--- a/VAS.Services/Controller/TemplatesController.cs
+++ b/VAS.Services/Controller/TemplatesController.cs
@@ -49,6 +49,7 @@ namespace VAS.Services.Controller
 	{
 		protected LimitationAsyncCommand<CreateEvent<TModel>> newTemplateCommand;
 		protected LimitationAsyncCommand<ImportEvent<TModel>> importTemplateCommand;
+		protected LimitationAsyncCommand<TModel> saveStaticTemplateCommand;
 
 		TemplatesManagerViewModel<TModel, TViewModel, TChildModel, TChildViewModel> viewModel;
 		ITemplateProvider<TModel> provider;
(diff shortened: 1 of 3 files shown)


VAS.Services/Controller/TemplatesController.cs
```diff
@@ -388,25 +391,18 @@ protected bool IsChildSelection (IEnumerable<TViewModel> selectedViewModels)
 				return;
 			}
 
-			if (template.Static) {
-				/* prompt=false when we click the save button */
-				if (force) {
-					evt.ReturnValue = await SaveStatic (template);
```
vguzmanp: Why not call `saveStaticTemplateCommand.ExecuteAsync (template)` here, instead of creating an extra event, etc?

VAS.Services/Controller/TemplatesController.cs
```diff
@@ -498,31 +499,27 @@ void HandleProviderCollectionChanged (object sender, NotifyCollectionChangedEven
 
 		#endregion
 
-		async Task<bool> SaveStatic (TModel template)
+		protected async virtual Task<bool> HandleSaveStatic (TModel template)
```
vguzmanp: Do you need it to be virtual?

VAS.Services/ViewModel/TemplateManagerViewModel.cs
```diff
@@ -43,6 +43,7 @@ public TemplatesManagerViewModel ()
 			LoadedTemplate = new TViewModel ();
 			NewCommand = new AsyncCommand (New);
 			SaveCommand = new AsyncCommand<bool> (Save, () => LoadedTemplate.Model != null && LoadedTemplate.Edited);
+			SaveStaticCommand = new AsyncCommand<bool> (Save, () => LoadedTemplate.Model != null && LoadedTemplate.Edited);
```
vguzmanp: Do we need this? It's the same as `SaveCommand`.

PR #407, opened 2017-10-02
LON-1151

SandraPG85:
Notify license change with the product code

### DIFF
--- a/VAS.Core/Events/Events.cs
+++ b/VAS.Core/Events/Events.cs
@@ -493,6 +493,7 @@ namespace VAS.Core.Events
 
 	public class LicenseChangeEvent : Event
 	{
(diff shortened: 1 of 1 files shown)


VAS.Core/Events/Events.cs
```diff
@@ -493,6 +493,7 @@ public class MoveElementsEvent<TParent, TChild> : Event
 
 	public class LicenseChangeEvent : Event
 	{
+		public uint ProductCode { get; set; }
```
vmartos: is this actually really needed? It is not necessary to check the product code, since this event is published every time a checklicense is performed. and the application should know if the license is valid by accessing App.Current.LicenseManager.LicenseStatus

PR #408, opened 2017-10-02
EventTypeLabelView should compute play button space

vmartos:
- in the RequiredWidth Property take into account playbutton space
- every time a FillCanvas is done changes WidthRequest to the max one

### DIFF
--- a/VAS.Drawing/Widgets/TimelineLabels.cs
+++ b/VAS.Drawing/Widgets/TimelineLabels.cs
@@ -118,6 +118,16 @@ namespace VAS.Drawing.Widgets
 				i++;
 			}
 			UpdateLabelOffsets ();
+			UpdateLabelsWidth ();
+		}
+
+		protected void UpdateLabelsWidth ()
+		{
+			double width = Objects.OfType<LabelView> ().Max (la => la.RequiredWidth);
+			foreach (LabelView lo in Objects.OfType<LabelView> ()) {
+				lo.Width = width;
+			}
+			WidthRequest = (int)width;
 		}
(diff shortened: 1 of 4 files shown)


VAS.Drawing/Widgets/TimelineLabels.cs
```diff
@@ -118,6 +118,16 @@ protected virtual void FillCanvas (ref int i)
...
 			}
 			UpdateLabelOffsets ();
+			UpdateLabelsWidth ();
+		}
+
+		protected void UpdateLabelsWidth ()
+		{
+			double width = Objects.OfType<LabelView> ().Max (la => la.RequiredWidth);
+			foreach (LabelView lo in Objects.OfType<LabelView> ()) {
+				lo.Width = width;
```
vguzmanp: Should each label have the width set, even if it's bigger than their required width? Shouldn't they just have the required, and let the container have the max?

VAS.UI.Gtk2/UI/TreeView/EventCellRenderer.cs
```diff
@@ -190,7 +190,7 @@ protected override void Render (Drawable window, Widget widget, Rectangle backgr
 				Area cell = new Area (new Point (cellArea.X, cellArea.Y),
 								cellArea.Width, cellArea.Height);
 
-				//Get thoffset to properly calulate if needs tooltip or redraw
+				//Get the ffset to properly calulate if needs tooltip or redraw
```
vguzmanp: lol, retypo'd

PR #409, opened 2017-10-03
Lon 993

vguzmanp:


VAS.Core/Common/ExtensionMethods.cs
```diff
@@ -261,5 +262,25 @@ public static bool Compare (this ExpandoObject obj1, ExpandoObject obj2)
 			}
 			return true;
 		}
+
+		/// <summary>
+		/// Checks if node1 is inside node2.
+		/// Inside means that any part of 1 is between start and stop of 2.
+		/// </summary>
+		public static bool IsInside (this TimeNode node1, TimeNode node2)
```
ylatuya: Extension Methods should only be used in my opinion for classes we can't extend ourselves. For example TimeNode already has some methods to Join and Intersect, the later looks similar to what you are trying to do.

VAS.Services/Controller/EventsFilterController.cs
```diff
@@ -0,0 +1,270 @@
...
+using Predicate = VAS.Core.Filters.Predicate<VAS.Core.ViewModel.TimelineEventVM>;
+
+namespace VAS.Services.Controller
+{
+	/// <summary>
+	/// Events filter controller.
+	/// This controller manages events filters and filtering for views where projects are used.
+	/// </summary>
+	public class EventsFilterController : ControllerBase<TimelineVM>
+	{
```
SandraPG85: Create a technical debt task to rethink the filters, the code is difficult to maintain and it is not possible to share code between them.

SandraPG85: Also in VAS.Tests the event filter controller should be tested, if it is in longomatch-core, move the test to VAS

SandraPG85: I think this class should be abstract, and the the update teams function as well

vguzmanp: https://fluendo.atlassian.net/browse/LON-1219

PR #412, opened 2017-10-06
LON-1213

SandraPG85:
Fix tag button tagging logic


VAS.Core/ViewModel/DashboardButtonVM.cs
```diff
@@ -574,6 +581,8 @@ public TagButtonVM ()
 			get;
 			set;
 		}
+
+		public Command TagClicked { get; set; }
```
ylatuya: Missing documentation. Command names should be imperative since they perform actions. Tag is redundant since the VM is already a TagButton and the action is to Click or to Toggle. I don't remind if we enforce the Command suffix in Command names

PR #413, opened 2017-10-06
Lon 1189

vguzmanp:


VAS.Core/App.cs
```diff
@@ -100,6 +100,7 @@ public static void Init (App appInit, string evUninstalled, string softwareName,
 			mainThreadId = Thread.CurrentThread.ManagedThreadId;
 
 			Current = appInit;
+			Current.SoftwareName = softwareName;
```
ylatuya: Is this still needed ?

vguzmanp: As none of my PR went in, yes
I cannot do anything without that.

When the first one goes in, the others will not have that commit

PR #423, opened 2017-10-17
Send TrackEvents for ShowUpgradeDialogs and UpgradeLinksClicked

vmartos:
- Two new events are created to publish when a upgrade dialog is shown and
  when user clicks an upgrade button
- UserStatisticsService subscribe to those events to be able to send TrackEvents
  to HockeyApp / AppInsights
- Add a GeneralProperties to send those properties to every trackEvent, use MergeProperties
  to merge general properties with custom properties for that event.

### DIFF
--- a/VAS.Core/Events/Events.cs
+++ b/VAS.Core/Events/Events.cs
@@ -504,4 +504,14 @@ namespace VAS.Core.Events
 	{
 		public TagButtonVM ButtonVM { get; set; }
 	}
+
+	public class LimitationDialogShownEvent
+	{
+		public string LimitationName { get; set; }
+		public string Source { get; set; }
+	}
+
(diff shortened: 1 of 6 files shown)


VAS.Core/Events/Events.cs
```diff
@@ -504,4 +504,14 @@ public class TagClickedEvent : Event
 	{
 		public TagButtonVM ButtonVM { get; set; }
 	}
+
+	public class LimitationDialogShownEvent
+	{
+		public string LimitationName { get; set; }
+		public string Source { get; set; }
+	}
+
```
SandraPG85: Keep the events separated to add the possibility of extension for both of them, instead of do inheritance.

PR #425, opened 2017-10-18
Add Image display feedback on vide player when a drawing is created

VoldarGames:
·[Implementation] Added new CollectionChanged Event Handler that loads last frame added/replaced in PlayList Drawings Collection

### DIFF
--- a/VAS.Services/VideoPlayerController.cs
+++ b/VAS.Services/VideoPlayerController.cs
@@ -53,26 +53,27 @@ namespace VAS.Services
 		public event MediaFileSetLoadedHandler MediaFileSetLoadedEvent;
 		public event PrepareViewHandler PrepareViewEvent;
 
-		protected const int TIMEOUT_MS = 20;
-		protected IVideoPlayer player;
-		protected IMultiVideoPlayer multiPlayer;
-		protected List<IViewPort> viewPorts;
-		protected ObservableCollection<CameraConfig> camerasConfig;
-		protected ObservableCollection<CameraConfig> defaultCamerasConfig;
-		protected object defaultCamerasLayout;
-		protected MediaFileSet defaultFileSet;
-		protected MediaFileSet mediafileSet;
-		protected MediaFileSet mediaFileSetCopy;
-		protected Time duration, videoTS, imageLoadedTS;
-		protected bool readyToSeek, stillimageLoaded, ready;
-		protected bool disposed, skipApplyCamerasConfig;
-		protected Action delayedOpen;
-		protected Seeker seeker;
-		protected Segment loadedSegment;
-		protected PendingSeek pendingSeek;
-		protected readonly Timer timer;
-		protected readonly ManualResetEvent TimerDisposed;
-		protected bool active;
+		const int TIMEOUT_MS = 20;
+
+		IVideoPlayer player;
+		IMultiVideoPlayer multiPlayer;
+		List<IViewPort> viewPorts;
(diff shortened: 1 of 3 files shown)


VAS.Services/VideoPlayerController.cs
```diff
@@ -1746,6 +1749,20 @@ void HandleLoadedTimelineEventPropertyChangedEventHandler (object sender, Proper
...
+		/// </summary>
+		/// <param name="sender">Sender.</param>
+		/// <param name="e">E.</param>
+		private void HandlePlaylistEventDrawingsCollectionChanged (object sender, NotifyCollectionChangedEventArgs e)
+		{
+			if ((e.Action == NotifyCollectionChangedAction.Add
+				|| e.Action == NotifyCollectionChangedAction.Replace)
+				&& e.NewItems.Count == 1) {
+				var editedFrameDrawing = e.NewItems [0] as FrameDrawing;
+				LoadPlayDrawing (editedFrameDrawing);
```
ylatuya: This doesn't check that the time matches the current time or that an event is loaded or that the new drawing matches the active camera. The Tick() function already has this logic in place except that maybe we need to change the following checks in  drawings.FirstOrDefault (f => f.Render > videoTS && f.Render <= currentTime) since at this point f.Render == videoTS

VAS.Services/VideoPlayerController.cs
```diff
@@ -1746,6 +1749,20 @@ void HandleLoadedTimelineEventPropertyChangedEventHandler (object sender, Proper
 				AbsoluteSeek (seekTime, true, false, true);
 			}
 		}
+		/// <summary>
+		/// Handles the playlist event drawings collection changed, drawing on the player the changed drawing frame
+		/// </summary>
+		/// <param name="sender">Sender.</param>
+		/// <param name="e">E.</param>
+		private void HandlePlaylistEventDrawingsCollectionChanged (object sender, NotifyCollectionChangedEventArgs e)
```
ylatuya: methods are private by default so according the coding style we don't add the private keyword

VAS.Services/VideoPlayerController.cs
```diff
@@ -1746,6 +1749,20 @@ void HandleLoadedTimelineEventPropertyChangedEventHandler (object sender, Proper
 				AbsoluteSeek (seekTime, true, false, true);
 			}
 		}
+		/// <summary>
+		/// Handles the playlist event drawings collection changed, drawing on the player the changed drawing frame
+		/// </summary>
+		/// <param name="sender">Sender.</param>
+		/// <param name="e">E.</param>
```
ylatuya: E alone doesn't provide any description.

VAS.Services/VideoPlayerController.cs
```diff
@@ -1746,6 +1749,20 @@ void HandleLoadedTimelineEventPropertyChangedEventHandler (object sender, Proper
 				AbsoluteSeek (seekTime, true, false, true);
 			}
 		}
+		/// <summary>
```
ylatuya: Missing line break between methods

VAS.UI.Gtk2/UI/Dialog/DrawingTool.cs
```diff
@@ -419,11 +419,11 @@ public void LoadFrame (Image frame)
 			UpdateTextSize ();
 		}
 
-		void Bind()
+		void Bind ()
```
ylatuya: These changes on this file are unrelated

PR #428, opened 2017-10-19
LON-1232

SandraPG85:
Fix do not use hockeyapp if not initialzed

### DIFF
--- a/VAS.Core/KPI/KpiService.cs
+++ b/VAS.Core/KPI/KpiService.cs
@@ -26,6 +26,8 @@ namespace VAS.KPI
 {
 	public class KpiService : IKpiService
 	{
+		bool initialized = false;
+
 		/// <summary>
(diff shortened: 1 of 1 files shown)


VAS.Core/KPI/KpiService.cs
```diff
@@ -53,28 +55,35 @@ public class KpiService : IKpiService
...
 			TrackEvent ("Session_Start", null, null);
+			initialized = true;
 #endif
 		}
 
 		public void TrackEvent (string eventName, IDictionary<string, string> properties = null, IDictionary<string, double> metrics = null)
 		{
 #if !DEBUG
-			HockeyClient.Current.TrackEvent (eventName, properties, metrics);
+			if (initialized) {
```
ylatuya: As per the coding style it's preferable to use a guard clause that checks preconditions and returns if it fails

PR #430, opened 2017-10-25
Set back ShowZoom to false after set to True

vmartos:


VAS.Core/ViewModel/VideoPlayerVM.cs
```diff
@@ -49,6 +49,7 @@ public VideoPlayerVM ()
 			ShowZoomCommand = new LimitationCommand (VASFeature.Zoom.ToString (),
 													 () => {
 														 ShowZoom = true;
+														 ShowZoom = false;
```
dfernandezramos: I think it's ok. But maybe for the developer is hard to understand why a property is set to true and immediately unseted. Wouldn't it be better to set the `ShowZoom` to false in the `PlayerVMPropertyChanged` method in  the VideoPlayerView after `zoomWindow.Show ();`?

vmartos: ok!

dfernandezramos:
👍

PR #436, opened 2017-11-03
Videocrashfix

ylatuya:


VAS.UI.Multimedia.Gtk2/UI/VideoPlayerView.cs
```diff
@@ -646,8 +646,16 @@ void OnVideoboxScrollEvent (object o, ScrollEventArgs args)
...
-			playerVM.ViewPorts = new List<IViewPort> { mainviewport, subviewport1.Viewport, subviewport2.Viewport, subviewport3.Viewport };
-			playerVM.ReadyCommand.Execute (true);
+			var viewPorts = new List<IViewPort> { mainviewport, subviewport1.Viewport, subviewport2.Viewport, subviewport3.Viewport };
+			playerVM.ViewPorts = viewPorts;
+			var activeCameras = playerVM.CamerasConfig.Count ();
+			// Check that all viewports are ready if cameras config are not set yet
+			// or check that we have at least the same number of viewports ready as the ones we
+			// need for the current cameras cofiguration.
+			if (!viewPorts.Any (vp => !(vp as VideoWindow).Ready) ||
+				(activeCameras != 0 && viewPorts.Take (activeCameras).Count (vp => (vp as VideoWindow).Ready) == activeCameras)) {
```
vguzmanp: Shouldn't it be `viewPorts.Count(...Ready) >= activeCameras`?

ylatuya: nope, if we have 2 cameras config we want the 2 first viewports ready, not the 3 and 4, that's why we use Take, to count only the first 2 and than check that all of them are ready

vguzmanp: Ok, the comment made me think that the important thing is the number, not the order.

vguzmanp: Also, now that I look at it, `!viewPorts.Any (vp => !(vp as VideoWindow).Ready)` is the same as `viewPorts.All vp => (vp as VideoWindow).Ready)`, right?

VAS.UI.Multimedia.Gtk2/UI/VideoPlayerView.cs
```diff
@@ -646,8 +642,21 @@ void OnVideoboxScrollEvent (object o, ScrollEventArgs args)
 
 		void HandleReady (object sender, EventArgs e)
 		{
-			playerVM.ViewPorts = new List<IViewPort> { mainviewport, subviewport1.Viewport, subviewport2.Viewport, subviewport3.Viewport };
-			playerVM.ReadyCommand.Execute (true);
+			bool AllViewPortsVisible (IEnumerable<IViewPort> vps)
+			{
+				return !vps.Any (vp => !(vp as VideoWindow).Ready);
```
vguzmanp: I don't really like local functions like this, but whatever

This should still be in positive, `All(Ready)` instead of `!Any(!Ready)`

PR #437, opened 2017-11-03
Improve OneWayPropertyBinding to be less verbose

ylatuya:
Add a new generic to type the target so that we don't have
to use a cast in the expression

### DIFF
--- a/VAS.Core/MVVMC/OneWayPropertyBinding.cs
+++ b/VAS.Core/MVVMC/OneWayPropertyBinding.cs
@@ -26,19 +26,20 @@ namespace VAS.Core.MVVMC
 	/// This class provides one way property binding.
 	/// <see cref="IViewModel" /> property is passed to write action callback performed on the class that setted the binding.
 	/// </summary>
-	public class OneWayPropertyBinding<T> : PropertyBinding<T>
+	public class OneWayPropertyBinding<TSourceProperty> : PropertyBinding<TSourceProperty>
 	{
-		Action<T> writeAction;
+		protected Action<TSourceProperty> writeAction;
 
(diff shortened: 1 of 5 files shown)


VAS.Core/MVVMC/OneWayPropertyBinding.cs
```diff
@@ -26,19 +26,20 @@ namespace VAS.Core.MVVMC
 	/// This class provides one way property binding.
 	/// <see cref="IViewModel" /> property is passed to write action callback performed on the class that setted the binding.
 	/// </summary>
-	public class OneWayPropertyBinding<T> : PropertyBinding<T>
+	public class OneWayPropertyBinding<TProperty> : PropertyBinding<TProperty>
```
VoldarGames: I think Property can be missunderstanding because it doesn't refer to an specific property source. I think TSourceProperty will be better.

VAS.Core/ViewModel/CountLimitationBarChartVM.cs
```diff
@@ -133,8 +133,7 @@ void UpdateVisibility ()
 
 		void HandlePropertyChanged (object sender, PropertyChangedEventArgs e)
 		{
-			if(Limitation == null || BarChart == null)
-			{
+			if (Limitation == null || BarChart == null) {
```
VoldarGames: I think this can be on a single line " if (Limitation == null || BarChart == null) return; "

ylatuya: That's just formatting code with was badly formatted

VAS.Core/MVVMC/OneWayPropertyBinding.cs
ylatuya: Ok, I'will change it

PR #438, opened 2017-11-06
Lma 146

vguzmanp:


VAS.Core/ViewModel/MenuNodeVM.cs
```diff
@@ -95,6 +95,11 @@ protected override void DisposeManagedResources ()
 			set;
 		}
 
+		public bool Selected {
+			get;
+			set;
+		}
+
 		/// <summary>
```
SandraPG85: Expose the menu icon property here and link it with the one provided by the command

VAS.Core/Common/ApplicationMenu.cs
```diff
@@ -0,0 +1,36 @@
...
+	/// For the moment this is used only in Mobile
+	/// </summary>
+	public class ApplicationMenu : IApplicationMenu
+	{
+		/// <summary>
+		/// Gets or sets the list of commands accesible from the application menu.
+		/// </summary>
+		/// <value>The possible commands.</value>
+		public IEnumerable<AsyncCommand> Tools { get; set; }
+	}
```
SandraPG85: Why you need the commands directly? You should have already a menuVm with the commands define there.

VAS.Core/StyleConf.cs
```diff
@@ -348,6 +348,7 @@ public StyleConf ()
 
 		public Color TBD_Black { get; set; } = Color.Black;
 		public Color TBD_Orange { get; set; } = Color.Orange;
+		public double SidebarBackgroundAlpha { get => 0.9; }
```
SandraPG85: The alpha should have a more generic name since it maybe reused by other components

VAS.Core/Interfaces/IApplicationMenu.cs
```diff
@@ -0,0 +1,29 @@
...
+using VAS.Core.Interfaces.MVVMC;
+using VAS.Core.ViewModel;
+
+namespace VAS.Core.Interfaces
+{
+	/// <summary>
+	/// Application global menu
+	/// </summary>
+	public interface IApplicationMenu : INestedViewModel<MenuNodeVM>
+	{ }
```
SandraPG85: the IApplicationMenu should expose the MenuVM to be used no?

PR #444, opened 2017-11-13
RA 870 - Modify drawing area tools behavior

axelgorris:
- Rename Rectangle area and Ellipsis area tools tooltips
- Remove unused tooltip properties in designer as they are set in code behind
- Do not draw borders when using drawing area tools
- Add option to specify color opacity. Defaults to 80
- Add FloatToUShort extension method


VAS.Core/Common/Color.cs
```diff
@@ -109,6 +109,11 @@ static public ushort ByteToUShort (byte val)
 			return (ushort)((float)val / byte.MaxValue * ushort.MaxValue);
 		}
 
+		static public ushort FloatToUShort (float val)
+		{
+			return Convert.ToUInt16 (ushort.MaxValue * val);
```
ylatuya: Can't you just cast here? (ushort)(ushort.MaxValue * val)

VAS.Drawing/Widgets/Blackboard.cs
```diff
@@ -473,7 +471,13 @@ protected override void StartMove (Selection sel)
 				if (copycolor) {
 					drawable.StrokeColor = Color.Copy ();
 				}
-				drawable.LineWidth = LineWidth;
+				if (Tool.Equals (DrawTool.RectangleArea) ||
+					Tool.Equals (DrawTool.CircleArea)) {
+					drawable.LineWidth = 0;
```
ylatuya: LineWidth is 0 by default, I would just change it when needed.

axelgorris: Done

VAS.Core/Common/Color.cs
axelgorris: Done

PR #445, opened 2017-11-14
Lon 1250

vguzmanp:


VAS.UI.Gtk2/UI/Common/TreeViewBase.cs
```diff
@@ -256,15 +256,24 @@ protected virtual void AddSubViewModel (IViewModel subViewModel, TreeIter parent
...
-			} else {
-				dictionaryStore [subViewModel].ForEach (x => store.Remove (ref x));
-				dictionaryStore.Remove (subViewModel);
+			if (subViewModel is INestedViewModel)
+			{
+				RemoveAllNestedSubViewModels(subViewModel);
+			}
+			else
+			{
+				TreeIter iter = dictionaryStore[subViewModel].First();
```
ylatuya: What happens for example with the players treeview in which an event can be in several players ?

vguzmanp: That was exactly the case that made this crash. The event was in each player's timeline, so each of them was calling remove, and the second one was crashing.
Now the first one will remove the first one, and the next calls will end up removing all of them.

VAS.UI.Gtk2/UI/Common/TreeViewBase.cs
```diff
@@ -263,8 +263,13 @@ protected virtual void RemoveSubViewModel (IViewModel subViewModel)
 			if (subViewModel is INestedViewModel) {
 				RemoveAllNestedSubViewModels (subViewModel);
 			} else {
-				dictionaryStore [subViewModel].ForEach (x => store.Remove (ref x));
-				dictionaryStore.Remove (subViewModel);
+				TreeIter iter = dictionaryStore [subViewModel].First ();
```
ylatuya: What happens in the players treeview where the same timelineeventvm is un several playervm? you will be removing just the first one, aren't you?

vguzmanp: (copypasted) That was exactly the case that made this crash. The event was in each player's timeline, so each of them was calling remove, and the second one was crashing.
Now the first one will remove the first one, and the next calls will end up removing all of them.

PR #446, opened 2017-11-14
Some fixes in Live Analysis

vmartos:
- Fix how CapturerBin handles Expose Events to avoid black screen in live.
- Fix close button event previewer.
- Fix edit event duration in Live Analysis by not displaying it in that mode.

### DIFF
--- a/VAS.UI.Multimedia.Gtk2/UI/CapturerBin.cs
+++ b/VAS.UI.Multimedia.Gtk2/UI/CapturerBin.cs
@@ -29,6 +29,7 @@ using VAS.Core.Store;
 using VAS.UI.Helpers;
 using Image = VAS.Core.Common.Image;
 using Misc = VAS.UI.Helpers.Misc;
+using Action = System.Action;
 
 namespace VAS.UI
 {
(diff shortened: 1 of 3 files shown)


VAS.UI.Multimedia.Gtk2/UI/CapturerBin.cs
```diff
@@ -353,15 +357,33 @@ protected virtual void HandleCloseClicked (object sender, EventArgs e)
...
 		public virtual void Run (CaptureSettings settings, MediaFile outputFile)
 		{
+			if (!capturerBinReady) {
+				delayedRun = () => InternalRun (settings, outputFile);
+				return;
+			}
+			InternalRun (settings, outputFile);
+		}
+
+		void InternalRun (CaptureSettings captureSettings, MediaFile outFile) {
```
vguzmanp: Shouldn't this be the virtual one?

VAS.UI.Multimedia.Gtk2/UI/VideoPlayerView.cs
```diff
@@ -746,37 +746,6 @@ void PlayerVMPropertyChanged (object sender, PropertyChangedEventArgs e)
 				editeventtimeruledrawingarea.Visible = ViewModel.EditEventDurationModeEnabled;
 				timerulearea.Visible = !ViewModel.EditEventDurationModeEnabled;
 				editDurationButton.Active = ViewModel.EditEventDurationModeEnabled;
-
-				//FIXME  We need Property binding like -> ctx.Add (control.Bind (c => c.Prop, vm => vm.Prop));
```
vguzmanp: I know we have the binding available now, but is this binded already?

VAS.UI.Multimedia.Gtk2/UI/CapturerBin.cs
vmartos: I don't know why it has protected fields and virtual methods since no one inherits from CapturerBin. But if someone inherits it should be the way you said, ill change it

VAS.UI.Multimedia.Gtk2/UI/VideoPlayerView.cs
vmartos: I removed this because it was duplicated, in fact it was also a method inside a method HandlePlayElementChanged inside PlayerVMPropertyChanged, since C# 7 admits this the compiler wasn't crashing. There was a mistake when rebasing somehow and thats why it was duplicated

VAS.UI.Multimedia.Gtk2/UI/CapturerBin.cs
vguzmanp: In that case just remove the protected & virtuals

VAS.UI.Multimedia.Gtk2/UI/VideoPlayerView.cs
vguzmanp: LOL! bad rebase that still compiles 😄

PR #447, opened 2017-11-14
RA-877 Make selection indicators more visible

axelgorris:
- Increase selection indicators size
- Add shadow to selection indicators
- Add 'DrawingSelectorAnchor', 'DrawingSelectorShadow',
'DrawingSelectorBorder' colors to theme
- Modify selection indicators colors

### DIFF
--- a/VAS.Core/StyleConf.cs
+++ b/VAS.Core/StyleConf.cs
@@ -176,6 +176,10 @@ namespace VAS.Core.Common
 		public static Color ButtonPenaltyColor = Color.Parse ("#ffc7f0");
 		public static Color ButtonEventColor = Color.Parse ("#c7e9ff");
 
+		public const int DrawingSelectorAnchorSize = 6;
+		public const int DrawingSelectorLineWidth = 1;
+		public const LineStyle DrawingSelectorLineStyle = LineStyle.Normal;
+
 		public int BenchLineWidth = 2;
 		public int TeamTaggerBenchBorder = 10;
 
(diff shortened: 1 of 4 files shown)


VAS.Drawing/CanvasObjects/CanvasObject.cs
```diff
@@ -277,11 +286,11 @@ protected void DrawSelectionArea (IDrawingToolkit tk)
...
 			}
-			tk.StrokeColor = Constants.SELECTION_INDICATOR_COLOR;
-			tk.StrokeColor = App.Current.Style.ThemeContrastDisabled;
+
+			tk.StrokeColor = App.Current.Style.DrawingSelectorBorder;
 			tk.FillColor = null;
-			tk.LineStyle = LineStyle.Dashed;
-			tk.LineWidth = 2;
+			tk.LineStyle = LineStyle.Normal;
+			tk.LineWidth = 1;
```
vguzmanp: Should this be hardcoded?

VAS.Core/StyleConf.cs
```diff
@@ -322,6 +322,9 @@ public StyleConf ()
 			PaletteTool = Color.Black;
 			TextBase = Color.Black;
 			SoftDim = 0.9;
+			DrawingSelectorAnchor = Color.White;
+			DrawingSelectorShadow = Color.Grey1;
+			DrawingSelectorBorder = Color.Grey1;
```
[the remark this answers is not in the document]
axelgorris: There are default values in case the are not set in the theme (same approach as HomeTeamColor for example)

VAS.Drawing/CanvasObjects/CanvasObject.cs
vguzmanp: yes, although the style will probably not be changed very often

PR #448, opened 2017-11-16
Lon 1255

vguzmanp:


VAS.Core/Interfaces/IPlaylistElement.cs
```diff
@@ -42,5 +43,21 @@ public interface IPlaylistElement : IChanged, INotifyPropertyChanged
 		Time Duration {
 			get;
 		}
+
+		/// <summary>
+		/// Gets or sets the cameras config.
+		/// </summary>
+		/// <value>The cameras config.</value>
+		ObservableCollection<CameraConfig> CamerasConfig {
```
ylatuya: I don't think that's correct, these properties are already in IPlaylistEventElement which are the only ones where CamerasConfig makes sense

vguzmanp: In that case, the problem we had will still happen with videos and images.
Right now `PlaylistVideo`, `PlaylistImage` and `PlaylistDrawing` are all defining a CamerasConfig of one camera, and their setter is private. How else can we update the camerasconfig of the player?

ylatuya: Ok, makes sense. In that case remove it from the IPlaylistEventElement which inherits from IPlaylistElement already

PR #449, opened 2017-11-16
LON-1254

dfernandezramos:


VAS.Services/Controller/TaggingController.cs
```diff
@@ -258,7 +258,7 @@ void HandleSubCategoriesTagging (AnalysisEventButtonVM buttonVM, TagVM subcatego
 
 		void HandleCapturerTick (CapturerTickEvent e)
 		{
-			project.Dashboard.CurrentTime = e.Time;
+			VideoPlayer.CurrentTime = e.Time;
```
vmartos: I Think this code should be the same way as it is was before, because the dashboard CurrentTime must be setted from the capturerTick when the loaded project is of type capture, so on the propertychanged here https://github.com/fluendo/VAS/blob/19f3e917c93fbfc3d6142d481c43b8cd796992cb/VAS.Services/Controller/TaggingController.cs#L226 should add an extra check that if we are in capture mode do not change project.Dashboard.CurrentTime.

Also the responsible for creating substitutions is the LMTeamTaggerController, this controller does not subscribe from CapturerTickEvents, so this controller must have the same, a subscription from the VideoPlayerVM.CurrentTime for project Files projects and a subscribtion from CapturerTickEvent when the loaded project is of type livecapture, fakeanalyisis, etc.

vmartos: Also could it be possible to add tests passing a project of type video and only changes the current time based on videoplayervm.CurrentTime and other test passing a project of type fakeanalysis and verify that the current time only changes from the capturertickevent

PR #450, opened 2017-11-16
Lma 154

VoldarGames:


VAS.Core/Events/GenericEvents.cs
```diff
@@ -24,6 +24,11 @@ public class GenericEvent<T> : ReturningValueEvent
 		public T Object { get; set; }
 	}
 
+	public class SearchEvent<T> : GenericEvent<T>
```
ylatuya: Why is it inheriting from the GenericEvent<T>? What do you need to pass as T Object ?

VAS.Core/Filters/VisibleRangeObservableProxy.cs
```diff
@@ -0,0 +1,110 @@
...
+					AddItem (item);
+				} else {
+					RemoveItem (item);
+				}
+			}
+			PropertyChangeList.Clear ();
+		}
+
+
+		List<TVisibleViewModel> PropertyChangeList = new List<TVisibleViewModel> ();
```
ylatuya: variable is not in the correct place and shouldn't start with capital letter

VAS.Core/Filters/VisibleRangeObservableProxy.cs
```diff
@@ -0,0 +1,110 @@
...
+
+		void HandlePropertyChanged (object sender, PropertyChangedEventArgs e)
+		{
+			if (e.PropertyName == nameof (IVisible.Visible)) {
```
ylatuya: This won't work when you change the visibility from True to False. You need to keep a list of items removed and added

VAS.Services/ViewModel/ProjectsManagerVM.cs
```diff
@@ -108,6 +107,12 @@ protected override void DisposeManagedResources ()
 			protected set;
 		}
 
+		[PropertyChanged.DoNotNotify]
+		public Command SearchProjectsCommand {
```
ylatuya: Missing documentation

VAS.Core/Filters/VisibleRangeObservableProxy.cs
```diff
@@ -0,0 +1,110 @@
...
+				var oldItems = e.OldItems.OfType<TVisibleViewModel> ();
+				foreach (var item in oldItems) {
+					RemoveItem (item);
+				}
+			}
+		}
+
+		void HandleCollectionChanged (object sender, NotifyCollectionChangedEventArgs e)
+		{
+			Update (sender, e);
```
ylatuya: The Update method is only called here, why do  you need a new function?

VAS.Core/Filters/VisibleRangeObservableProxy.cs
```diff
@@ -0,0 +1,110 @@
...
+		}
+
+		public VisibleRangeObservableProxy ()
+		{
+		}
+
+		void AddItem (TVisibleViewModel viewModel)
+		{
+			if (viewModel.Visible) {
+				viewModel.PropertyChanged += HandlePropertyChanged;
```
ylatuya: You need to monitor all objects, not just the visible ones. If you remove and add the same event you will have it connected twice

VAS.Tests/Core/Filters/TestVisibleRangeObservableProxy.cs
```diff
@@ -0,0 +1,101 @@
...
+
+				set {
+					visible = value;
+					OnPropertyChanged (new System.ComponentModel.PropertyChangedEventArgs (nameof (Visible)));
+				}
+			}
+		}
+
+		[Test ()]
+		public void AddItems_SomeVisiblesOthersNot_GetsOnlyVisibleItems ()
```
ylatuya: You will have to write some more tests, there are many scenarios you are not considering

VAS.Core/Filters/VisibleRangeObservableProxy.cs
ylatuya: Also, this is already correctly implemented in the BindableBase, reuse it please

VAS.Core/Events/GenericEvents.cs
VoldarGames: Now is used for SearchEvent<LMProject> but in future could be for other types.
Like the other ones below this line.

VAS.Services/ViewModel/ProjectsManagerVM.cs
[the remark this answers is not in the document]
VoldarGames: I don't understand your comment. 
It is named Command SearchProjectsCommand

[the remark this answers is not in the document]
vmartos: yep the original comment was that should be named 'Command<string> SearchProjectsCommand', since the parameter accepted by this command should be always a string.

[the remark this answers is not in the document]
VoldarGames: The parameter is not only string, theres a case that it is an EventArg

VAS.Core/Filters/VisibleRangeObservableProxy.cs
```diff
@@ -0,0 +1,111 @@
...
+			if (e.Action == NotifyCollectionChangedAction.Add)
+			{
+				var newItems = e.NewItems.OfType<TVisibleViewModel> ().Where (tVisibleViewModel => tVisibleViewModel.Visible);
+				foreach (var item in newItems)
+				{
+					item.PropertyChanged += HandlePropertyChanged;
+					AddItem (item);
+				}
+			}
+			else if (e.Action == NotifyCollectionChangedAction.Remove)
```
vmartos: What will happen if we actually. Clear the list? or Reset the list? I think you should care also for NotifyCollectionChangedAction.Replace:
NotifyCollectionChangedAction.Reset:

cases, Add a test to check what happens if we replace a collection with another one

VAS.Core/Filters/VisibleRangeObservableProxy.cs
```diff
@@ -0,0 +1,111 @@
...
+using System.Collections.Specialized;
+using System.ComponentModel;
+using System.Linq;
+using VAS.Core.Common;
+using VAS.Core.Interfaces.GUI;
+using VAS.Core.MVVMC;
+
+namespace VAS.Core.Filters
+{
+	public class VisibleRangeObservableProxy<TVisibleViewModel> : RangeObservableCollection<TVisibleViewModel> where TVisibleViewModel : ViewModelBase, IVisible
```
vmartos: Document class

PR #454, opened 2017-11-19
Improve preformance of filters

ylatuya:
In composite predicates emit the Active property chagned
event only once

### DIFF
--- a/VAS.Core/Filters/Predicate.cs
+++ b/VAS.Core/Filters/Predicate.cs
@@ -19,7 +19,6 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
 using System.ComponentModel;
 using System.Linq;
 using System.Linq.Expressions;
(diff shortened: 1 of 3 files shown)


VAS.Core/Filters/Predicate.cs
```diff
@@ -130,9 +129,14 @@ protected override void DisposeManagedResources ()
 			}
 
 			set {
+				IgnoreEvents = true;
+				bool emit = false;
 				foreach (var item in Elements) {
+					emit |= item.Active != value;
```
vguzmanp: Can we use `IsChanged` for this? In theory Fody is already doing that when changed, right?

vguzmanp:
You are basing the pull request on longomatch-release when it should be based on master.

VAS.Core/Filters/Predicate.cs
ylatuya: IsChanged is only changed back to false when we save.
I also forgot to remove that since in composite filters this won't work if you just change one of the childs, we should just emit always

PR #455, opened 2017-11-20
LMA-150

SandraPG85:
Add card menu to send a project to the trash

### DIFF
--- a/VAS.Core/ViewModel/MenuNodeVM.cs
+++ b/VAS.Core/ViewModel/MenuNodeVM.cs
@@ -18,6 +18,7 @@
 using System;
 using System.Linq;
 using VAS.Core.MVVMC;
+using VAS.Core.Common;
 
 namespace VAS.Core.ViewModel
 {
@@ -76,6 +77,12 @@ namespace VAS.Core.ViewModel
 			get;
 			private set;
 		}
+
+		/// <summary>
+		/// Gets the color of the menu item.
+		/// </summary>
+		/// <value>The color of the menu item.</value>
+		public Color Color { get; set; }
 
 		/// <summary>
 		/// Gets the submenu.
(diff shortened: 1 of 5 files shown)


VAS.Core/ViewModel/MenuNodeVM.cs
```diff
@@ -77,6 +78,12 @@ protected override void DisposeManagedResources ()
 			private set;
 		}
 
+		/// <summary>
+		/// Gets the color of the menu item.
+		/// </summary>
+		/// <value>The color of the menu item.</value>
+		public Color ItemColor { get; set; }
```
vguzmanp: Shouldn't it be just `Color`?

VAS.Services/ViewModel/ProjectsManagerVM.cs
```diff
@@ -41,6 +41,7 @@ public ProjectsManagerVM ()
 			DeleteCommand = new AsyncCommand (Delete, () => Selection.Any ());
 			SaveCommand = new AsyncCommand (Save, () => LoadedProject?.Model != null && LoadedProject.IsChanged);
 			ExportCommand = new AsyncCommand (Export, () => Selection.Count == 1);
+			TrashCommand = new AsyncCommand<TViewModel> (Trash, () => true) { IconName = "lma-trash" };
```
vguzmanp: Why is this a different command than Delete?

SandraPG85: the delete operation works with the selection and allows the multiple project deletion, instead this one only removes the project linked to the trash command

vguzmanp: But in the future we will have multiselection for deletion too, we should just use the same command and create the selection if it's only one.

PR #462, opened 2017-11-21
Limit the ROI in the CamerasConfig Setter instead of the Getter

vmartos:


VAS.Services/VideoPlayerController.cs
```diff
@@ -155,6 +155,18 @@ protected override void DisposeManagedResources ()
 			set {
 				Log.Debug ("Updating cameras configuration: " + value);
 				camerasConfig = value;
+
+				//Limit Region of interest if Open Zoom for Events/Filesets is limited
+				if (camerasConfig != null) {
+					if (!App.Current.LicenseLimitationsService.CanExecute
+						(VASFeature.OpenZoom.ToString ())) {
+						camerasConfig = camerasConfig.Clone ();
```
ylatuya: I think Clone() might have been used at some point with multicameras. Have you check that it still works correctly?

VAS.Services/VideoPlayerController.cs
```diff
@@ -756,7 +762,10 @@ public virtual void Previous (bool force = false)
 
 		public virtual void ApplyROI (CameraConfig camConfig)
 		{
-			camerasConfig [camConfig.Index] = camConfig;
+
+			int index = camerasConfig.IndexOf (camConfig);
```
ylatuya: I am not sure to understand this change

vmartos: The problem I was facing here, and i think this problem is also on LM 1.3.7, were that camConfig.Index is not the same as the index inside camerasConfig, so when trying to zoom in a camera with index 1 but was the main camera now, it was setting the camerasConfig [1] equal to camerasConfig [0] because at that moment camerasConfig[1] had index 1

PR #464, opened 2017-11-21
Vas 93 fix

VoldarGames:


VAS.Services/VideoPlayerController.cs
```diff
@@ -1548,10 +1548,13 @@ bool Tick (Time currentTime = null)
 							if (drawings != null) {
 								/* Check if the event has drawings to display */
 								FrameDrawing fd = drawings.FirstOrDefault (f => IsDrawingVisibleForCurrentTime (f, currentTime));
-								if (fd != null) {
-									LoadPlayDrawing (fd);
+								if (fd != null && fd != playerVM.FrameDrawing) {
```
ylatuya: Why not moving all these checks to the IsDrawingVisibleForCurrentTime

VAS.Services/VideoPlayerController.cs
```diff
@@ -1547,10 +1547,10 @@ bool Tick (Time currentTime = null)
 							var drawings = LoadedTimelineEvent?.Drawings;
 							if (drawings != null) {
 								/* Check if the event has drawings to display */
-								FrameDrawing fd = drawings.FirstOrDefault (f => IsDrawingVisibleForCurrentTime (f, currentTime));
-								if (fd != null) {
-									LoadPlayDrawing (fd);
-								}
+								foreach (var drawing in drawings)
```
ylatuya: This will iterate over all drawings instead of just drawing the first one that matches the condition

PR #467, opened 2017-11-22
Fix tool selection after creating an element

ylatuya:


VAS.UI.Gtk2/UI/Dialog/DrawingTool.cs
```diff
@@ -617,8 +636,23 @@ protected override bool OnKeyPressEvent (Gdk.EventKey evnt)
 
 		void HandleToolClicked (object sender, EventArgs e)
 		{
-			lastButton = sender as RadioButton;
-			blackboard.Tool = GetTool (lastButton);
+			if (sender is RadioButton button) {
```
dfernandezramos: Cool! Didn't know this C# feature :D

ylatuya: new in C# 7 I think :D

PR #468, opened 2017-11-23
Fix AboutDialog StateMoveBack OnDestroyed

VoldarGames:


VAS.UI.Gtk2/UI/Dialog/AboutDialog.cs
```diff
@@ -86,6 +86,12 @@ public void OnUnload ()
 
 		}
 
+		protected override void OnDestroyed ()
+		{
+			App.Current.GUIToolkit.Invoke (async (s, e) => await App.Current.StateController.MoveBack ());
```
ylatuya: You don't need to use App.Current.GUIToolkit.Invoke here.
You should use the DeleteEvent instead of the Destroy event here. Destroy is already too late since the dialog is already destroyed and the state controller can't for instance cancel the transition.

ylatuya:
You should include here the fix for the license width being to big

VoldarGames:
answering "You should include here the fix for the license width being to big", it is fixed on riftanalyst bitbucket repository

VAS.UI.Gtk2/UI/Dialog/AboutDialog.cs
VoldarGames: bool OnDeleteEvent (Gdk.Event evnt) is not being called after window close. Can I use OnDestroyed? it is working with this event, of course I will remove "App.Current.GUIToolkit.Invoke" if it is not necessary.

PR #471, opened 2017-11-23
Fixes SubViewPorts appearing when it shouldn't

vmartos:


VAS.UI.Multimedia.Gtk2/UI/VideoPlayerView.cs
```diff
@@ -749,7 +749,7 @@ void PlayerVMPropertyChanged (object sender, PropertyChangedEventArgs e)
 			}
 			//FIXME  We need Property binding like -> ctx.Add (control.Bind (c => c.Prop, vm => vm.Prop));
 			if (ViewModel.NeedsSync (e, nameof (ViewModel.ViewPortsSwitchActive))) {
-				subviewportsbox.Visible = ViewModel.ViewPortsSwitchActive;
+				subviewportsbox.Visible = ViewModel.ViewPortsSwitchActive && (ViewModel.FileSet != null && ViewModel.FileSet.ViewModels.Count > 1);
```
vguzmanp: Can't you set it with `SubViewPortsVisible` here?

ylatuya: I also proposed the same the other day when we discussed since the logic is already there

PR #472, opened 2017-11-23
Add version to IDevice and set it from App.Init

vguzmanp:


VAS.Core/App.cs
```diff
@@ -216,6 +216,9 @@ static void InitTranslations (string softwareName)
 
 		static void InitVersion ()
 		{
+#if OSTYPE_ANDROID || OSTYPE_IOS
```
ylatuya: I don't think you need this, do you? The Desktop application must also implement IDevice.Version and move this code the IDevice.Version implementation

PR #476, opened 2017-11-24
LON-1286

dfernandezramos:


VAS.Core/ViewModel/DashboardButtonVM.cs
```diff
@@ -719,6 +718,10 @@ public bool Start (List<DashboardButtonVM> buttons)
 		/// <param name="buttons">Buttons.</param>
 		public void Stop (List<DashboardButtonVM> buttons)
 		{
+			if (currentNode == null) {
+				return;
+			}
+
```
SandraPG85: is it possible to add a quick test to ensure that no exception happens if current node is null?

PR #483, opened 2017-11-29
Add support for default values in bindings

VoldarGames:


VAS.Core/MVVMC/OneWayPropertyBinding.cs
```diff
@@ -28,11 +28,13 @@ namespace VAS.Core.MVVMC
 	/// </summary>
 	public class OneWayPropertyBinding<TSourceProperty, TTargetProperty> : PropertyBinding<TSourceProperty, TTargetProperty>
 	{
+		object defaultValue;
```
ylatuya: defaultValues should be typed with the generic

VAS.Core/MVVMC/OneWayPropertyBinding.cs
```diff
@@ -53,7 +55,14 @@ protected override void UnbindView ()
 
 		protected override void WriteViewValue (TTargetProperty val)
 		{
-			this.writeAction (val);
+			if (Equals (val, default (TTargetProperty))) {
+				try {
+					val = (TTargetProperty)defaultValue;
```
ylatuya: This cast for the defaultValue shouldn't be needed

VAS.Core/MVVMC/OneWayPropertyBinding.cs
```diff
@@ -53,7 +55,14 @@ protected override void UnbindView ()
 
 		protected override void WriteViewValue (TTargetProperty val)
 		{
-			this.writeAction (val);
+			if (Equals (val, default (TTargetProperty))) {
```
ylatuya: This will call the Equals method of the current type, not val or TTargetProperty

VoldarGames: I paste from documentation about object.Equals method : "determines whether the specified object instances are considered equal". 
I can't use == operator because TTargetProperty doesn't have class() constraint and TTargetType could be a ValueType such as int.

VAS.Core/MVVMC/OneWayPropertyBinding.cs
VoldarGames: Cast is not needed here but would be on "his.defaultValue = defaultValue;" at ctor()

VAS.Core/MVVMC/OneWayPropertyBinding.cs
```diff
@@ -53,7 +58,14 @@ protected override void UnbindView ()
 
 		protected override void WriteViewValue (TTargetProperty val)
 		{
-			this.writeAction (val);
+			if (object.Equals (val, default (TTargetProperty))) {
+				try {
```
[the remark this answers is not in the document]
ylatuya: Also, do you really need this assignment ?

PR #484, opened 2017-11-29
Lon 1288

dfernandezramos:


VAS.Core/Store/Tag.cs
```diff
@@ -58,15 +58,6 @@ public override bool Equals (object obj)
 			return Value == tag.Value && Group == tag.Group;
 		}
```
SandraPG85: Add a comment to re add the gethashcode in the future using inmutables properties instead

PR #486, opened 2017-11-29
Don't store the TaskCompletionSource in the class

vguzmanp:
This fixes the 'System.InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed.' exception
because a popup was pushed before the last one had been able to set its result.

### DIFF
--- a/VAS.UI.Gtk2/GUIToolkitBase.cs
+++ b/VAS.UI.Gtk2/GUIToolkitBase.cs
@@ -40,8 +40,6 @@ namespace VAS.UI
 	{
 		Gtk.Window mainWindow;
 		Registry registry;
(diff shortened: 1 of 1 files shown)


VAS.UI.Gtk2/GUIToolkitBase.cs
```diff
@@ -290,9 +289,8 @@ public Task PopModal (IPanel panel)
 			return AsyncHelpers.Return ();
 		}
 
-		protected void ShowModalWindow (IPanel panel, IPanel parent)
+		protected bool ShowModalWindow (IPanel panel, IPanel parent)
 		{
-			taskResult = true;
```
ylatuya: Isn't taskResult also a single instance variable that can be overriden?

PR #488, opened 2017-11-30
Lma 152

ylatuya:


VAS.Core/StyleConf.cs
```diff
@@ -98,6 +94,10 @@ public StyleConf ()
 		public Color ColorAccentSuccess { get; set; }
 		public Color ColorAccentWarning { get; set; }
 		public Color ColorAccentError { get; set; }
+		public Color ColorFullWhite { get; set; }
+		public Color ColorShadow { get; set; }
+
+		public Color DrawingSelectorBorder { get; set; }
```
vguzmanp: All start with `Color` except for this one

VAS.Core/StyleConf.cs
```diff
@@ -98,6 +94,10 @@ public StyleConf ()
 		public Color ColorAccentSuccess { get; set; }
 		public Color ColorAccentWarning { get; set; }
 		public Color ColorAccentError { get; set; }
+		public Color ColorFullWhite { get; set; }
```
vguzmanp: This shouldn't include the color name.

ylatuya: It must include it because it's an absolute white, independent of the theme

VAS.Core/StyleConf.cs
ylatuya: This color was already in the palette, I have left it as it is

VAS.Core/StyleConf.cs
vguzmanp: Then it doesn't make sense to have it in the palette.

Why do we need absolute white, independent of the theme?

PR #494, opened 2017-12-11
LON-1288

SandraPG85:
Do a collection that update the key when a view model property changes

### DIFF
--- a/VAS.Core/MVVMC/KeyUpdaterCollectionViewModel.cs
+++ b/VAS.Core/MVVMC/KeyUpdaterCollectionViewModel.cs
@@ -1,0 +1,109 @@
+﻿//
+//  Copyright (C) 2017 
+//
+//  This program is free software; you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation; either version 2 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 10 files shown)


VAS.Core/MVVMC/KeyUpdaterCollectionViewModel.cs
```diff
@@ -0,0 +1,62 @@
...
+		}
+
+		protected override TViewModel CreateInstance (TModel model)
+		{
+			var vm = base.CreateInstance (model);
+			vm.PropertyChanged += HandlePropertyChanged;
+			return vm;
+		}
+
+		void HandlePropertyChanged (object sender, System.ComponentModel.PropertyChangedEventArgs e)
```
ylatuya: This seems to aggressive, it will recreate the dictionary on each property changed of the view model, even if it's not changing the hashcode

VAS.Core/MVVMC/KeyUpdaterCollectionViewModel.cs
```diff
@@ -0,0 +1,69 @@
...
+				vm.PropertyChanged -= HandlePropertyChanged;
+			}
+
+			base.DisposeUnmanagedResources ();
+		}
+
+		protected override TViewModel CreateInstance (TModel model)
+		{
+			var vm = base.CreateInstance (model);
+			vm.PropertyChanged += HandlePropertyChanged;
```
ylatuya: There is a problem doing it here, since you only connect the property change but you don't disconnect when the value is removed

PR #495, opened 2017-12-11
Fix Tagging Manual EventTypes with Hotkeys

VoldarGames:


ylatuya:
It's hard to know how you fixed it because you have split files and the fix in the same commit so I don't really know what to look for

PR #498, opened 2017-12-12
LON-1311 Bug Fix Image in top of playlist doesn't show with double click

VoldarGames:


VAS.UI.Multimedia.Gtk2/UI/VideoPlayerView.cs
```diff
@@ -755,6 +756,11 @@ void PlayerVMPropertyChanged (object sender, PropertyChangedEventArgs e)
 			}
 		}
 
+		bool IsPlaylistImageLoaded ()
```
ylatuya: This function is mixed the the callback functions.

VAS.UI.Multimedia.Gtk2/UI/VideoPlayerView.cs
```diff
@@ -755,6 +756,11 @@ void PlayerVMPropertyChanged (object sender, PropertyChangedEventArgs e)
...
+		{
+			return (ViewModel.LoadedElement != null && ViewModel.LoadedElement is PlaylistImage);
```
ylatuya: Do you really need a function for that? In case you want to keep it, you can make it a propety and move it upper since it's not mixed with callback functions.
The first check is not need since when LoadedElement is null ViewModel.LoadedElement is PlaylistImage will return also false

VAS.UI.Multimedia.Gtk2/UI/VideoPlayerView.cs
```diff
@@ -729,8 +741,9 @@ void PlayerVMPropertyChanged (object sender, PropertyChangedEventArgs e)
 			if (ViewModel.NeedsSync (e, nameof (ViewModel.FrameDrawing))) {
 				if (playerVM.FrameDrawing != null) {
 					LoadImage (playerVM.CurrentFrame, playerVM.FrameDrawing);
+					DrawingsVisible = true;
```
ylatuya: LoadImage is already setting DrawingsVisible = true

VAS.UI.Multimedia.Gtk2/UI/VideoPlayerView.cs
```diff
@@ -697,11 +703,17 @@ void PlayerVMPropertyChanged (object sender, PropertyChangedEventArgs e)
 			}
 			if (ViewModel.NeedsSync (e, nameof (ViewModel.LoadedElement))) {
 				HandlePlayElementChanged ();
+				if (ViewModel.LoadedElement == null) {
+					DrawingsVisible = false;
+				} else {
+					ViewModel.FrameDrawing = null;
```
ylatuya: The view can't set this property to null, only the controller can

VAS.Services/VideoPlayerController.cs
```diff
@@ -972,6 +972,7 @@ void EmitElementLoaded (IPlaylistElement element, Playlist playlist)
 		{
 			playerVM.NextCommand.Executable = playlist != null ? playlist.HasNext () : false;
 			playerVM.LoadedElement = element;
+			if (element != null) playerVM.FrameDrawing = null;
```
ylatuya: There is already en EmitLoadDrawings that you should call with EmitLoadDrawings (null); if you want to unload a potential frame drawing as it's done in other places in the code EmitElementLoaded is not the place to do it.
First write a test with that reproduces your issue and than write the fix.

VAS.Services/VideoPlayerController.cs
```diff
@@ -623,6 +623,7 @@ public void Switch (TimelineEvent play, Playlist playlist, IPlaylistElement elem
 
 			loadedEvent = play;
 			loadedPlaylistElement = element;
+			if (element != null) playerVM.FrameDrawing = null;
```
ylatuya: I think this should always be reset, even if element is null which is when element is unloaded

PR #500, opened 2017-12-12
Ask for confirmation when deleting events

vmartos:


VAS.Services/Controller/EventsController.cs
```diff
@@ -309,16 +309,24 @@ void AddNewPlay (TimelineEvent play)
 			}
 		}
 
-		void DeletePlays (IEnumerable<TimelineEvent> plays, bool update = true)
+		async Task DeletePlays (IEnumerable<TimelineEvent> plays, bool askConfirmation = true)
 		{
 			plays = plays.Where (p => p.Deletable);
 			Log.Debug (plays.Count () + " plays deleted");
+			if (askConfirmation) {
+				var delete = await App.Current.Dialogs.QuestionMessage (
```
ylatuya: I think we shouldn't use {plays.Count()} for translations, as it would introduce translation bugs

PR #502, opened 2017-12-12
Track exceptions also in other rendering errors

vguzmanp:


VAS.Core/Common/Exceptions.cs
```diff
@@ -142,4 +142,9 @@ private static string GenerateMessage (string name)
 			}
 		}
 	}
+
```
VoldarGames: Missing Summary

PR #504, opened 2017-12-13
Circle resizing is done from opposite anchor of the element

axelgorris:


VAS.Core/Store/Drawables/Ellipse.cs
```diff
@@ -22,79 +22,89 @@
...
+			set {
+				axisY = value;
+				Update ();
+			}
 		}
 
 		[JsonIgnore]
 		[PropertyChanged.DoNotNotify]
-		public Point Top {
+		public override Point TopLeft {
```
ylatuya: just as a recommendation you can use the new syntax:
public override Point TopLeft => base.TopLeft

VAS.Core/Store/Drawables/Ellipse.cs
```diff
@@ -22,79 +22,89 @@
...
 			AxisY = axisY;
+			ignoreChanges = false;
+			Update ();
 			Text = text;
 		}
 
-		public Point Center {
-			get;
-			set;
+		Point center;
```
ylatuya: Move local variables to the correct place

VAS.Drawing.Cairo/CairoBackend.cs
```diff
@@ -485,6 +485,9 @@ public void DrawCircleImage (Point center, double radius, Image image)
 
 		public void DrawEllipse (Point center, double axisX, double axisY)
 		{
+			axisX = Math.Abs (axisX);
```
ylatuya: AxisY and AxisX shouldn't be negative. The problem looks more on the Ellipse side

VAS.Core/Store/Drawables/Ellipse.cs
```diff
@@ -22,79 +22,89 @@
...
+		public new Point Center {
```
ylatuya: Center is virtual, shouldn't you use an override here?

VAS.Core/Store/Drawables/Ellipse.cs
```diff
@@ -107,76 +100,67 @@ public Ellipse (Point center, double axisX, double axisY, string text = null)
...
+
 		public override Selection GetSelection (Point p, double pr = 0.05, bool inMotion = false)
 		{
-			double d;
+			var selection = base.GetSelection (p, pr);
 
-			if (Selected) {
-				return base.GetSelection (p, pr);
-			}
+			if (selection != null)
```
ylatuya: if Selection != null and Selection.Position != SelectionPosition.All

Otherwise you will never enter in the next section that checks if the click is inside the ellipse

PR #507, opened 2017-12-18
Lon 1325

ylatuya:


VAS.Services/Controller/EventsFilterController.cs
```diff
@@ -227,34 +223,37 @@ protected virtual void UpdateEventTypesPredicates (object sender = null, NotifyC
 				var analysisEventType = eventType.Model as AnalysisEventType;
 				if (analysisEventType != null && analysisEventType.Tags.Any ()) {
 					CompositePredicate<TimelineEventVM> composedEventTypePredicate;
+					var composedPredicates = new List<IPredicate<TimelineEventVM>> (); ;
```
vguzmanp: Extra `;`

VAS.Tests/Services/TestEventsFilterController.cs
```diff
@@ -69,6 +69,47 @@ public void OneTimeSetUp ()
...
+		{
+			// Arrange
+			// Arrange
+			try {
+				await eventsFilterController.Stop ();
+			} catch (InvalidOperationException) {
+				// Ignore the already stopped error
+			}
+			timelineVM.EventsPredicate.Clear ();
+			timelineVM.FiltersInitialized = true;
```
vguzmanp: I don't really like being able to do this, but I don't see other way :/

ylatuya: yeah, this could always happen if someone is using incorrectly the API :)

PR #510, opened 2017-12-28
RA-1111

vmartos:


VAS.Services/DynamicButtonToolbarService.cs
```diff
@@ -0,0 +1,68 @@
...
+using System;
+using System.Threading.Tasks;
+using VAS.Core.Events;
+using VAS.Core.Interfaces;
+using VAS.Core.Interfaces.MVVMC;
+using VAS.Core.MVVMC;
+using VAS.Core.ViewModel;
+
+namespace VAS.Services
+{
```
SandraPG85: comment class in general

VAS.Services/DynamicButtonToolbarService.cs
```diff
@@ -0,0 +1,68 @@
...
+	public class DynamicButtonToolbarService : ControllerBase<DynamicButtonToolbarVM>, IService
+    {
```
SandraPG85: Longomatch has a similar ToolsService, maybe we can unify the services to register commands used in external widgets. So we can have a unique commands repository and the his in the load state access to it.
In longomatch you don't have to do it now just remember to create the tech debt task, but a more generic service maybe with a categorised registration could be helpful.

VAS.Services/ViewModel/ProjectsManagerVM.cs
```diff
@@ -39,7 +39,7 @@ public class ProjectsManagerVM<TModel, TViewModel> : LimitedCollectionViewModel<
 		public ProjectsManagerVM ()
 		{
 			LoadedProject = new TViewModel ();
-			NewCommand = new Command (New) { IconName = "vas-plus" };
+			NewCommand = new LimitationCommand (VASCountLimitedObjects.Projects.ToString (), New) { IconName = "vas-plus" };
 			OpenCommand = new AsyncCommand<TViewModel> (Open, (arg) => Selection.Count == 1);
```
SandraPG85: Add a test for the bug that you commented in the daily,  about creating new projects without limitation

VAS.Services/DynamicButtonToolbarService.cs
vmartos: the ToolsService in LongomatchPro is used to register the PresentationManagerState and to add the PresentationManager ITool implementation to the list of ITools. So is a total different purpose and to have different services for different purposes is totally legit. I prefer different unit testable services than a Service that manages too much things that are not related.

VAS.Core/Interfaces/GUI/IScreenState.cs
```diff
@@ -53,6 +54,12 @@ public interface IScreenState : IDisposable
 		/// <value>The key context.</value>
 		KeyContext KeyContext { get; }
 
+		/// <summary>
+		/// Gets the toolbar commands.
+		/// </summary>
+		/// <value>The toolbar commands.</value>
+		IEnumerable<Command> ToolbarCommands { get; }
+
```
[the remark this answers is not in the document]
vmartos: created: https://fluendo.atlassian.net/browse/VAS-175

VAS.Services/State/ScreenState.cs
```diff
@@ -89,6 +89,8 @@ protected override void DisposeManagedResources ()
 			protected set;
 		}
 
+		public virtual IEnumerable<Command> ToolbarCommands => new List<Command> ();
+
```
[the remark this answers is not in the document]
SandraPG85: comment the public property

PR #512, opened 2018-01-03
Add long to string converter

vmartos:


VAS.Core/Common/TypeConverters/LongToStringConverter.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.ComponentModel;
+using System.Globalization;
+
+namespace VAS.Core.Common.TypeConverters
+{
+	public class LongToStringConverter: TypeConverter
+	{
+		public override bool CanConvertTo (ITypeDescriptorContext context, Type destinationType)
+		{
+			return (destinationType == typeof (long));
```
vguzmanp: Is this swapped with `CanConvertFrom`? the `destinationType` should be string, shouldn't it?

vmartos: I will remove this class since it is not necessary anymore, the documentation on this is quite strange:

https://msdn.microsoft.com/en-us/library/ayybcxe5.aspx

This PointConverter is an example, the CanConvertFrom is of type string... I don't know why it does not override the CanConvertTo...

Also the TimeToStringConverter is similar, the CanConvertTo checks destination type is Time and CanConvertFrom checks string...

PR #514, opened 2018-01-05
RA-1195

SandraPG85:
Add shape support to the image view

### DIFF
--- a/VAS.UI.Helpers.Gtk2/ImageView.cs
+++ b/VAS.UI.Helpers.Gtk2/ImageView.cs
@@ -110,6 +110,24 @@ namespace VAS.UI.Helpers
 		}
 
 		/// <summary>
+		/// Gets or sets a value indicating whether this <see cref="T:VAS.UI.Helpers.ImageView"/> is rendered in a circular shape.
+		/// </summary>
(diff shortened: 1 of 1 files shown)


VAS.UI.Helpers.Gtk2/ImageView.cs
```diff
@@ -193,7 +211,26 @@ protected override bool OnExposeEvent (EventExpose evnt)
...
+
+					Point center = new Point (r.X + r.Width / 2, r.Y + r.Height / 2);
+					double radius = Math.Min (height, width) / 2;
+
+					// Apply shadow
+					if (HasShadow)
+					{
+						App.Current.DrawingToolkit.FillColor = ShadowColor;
+						App.Current.DrawingToolkit.LineWidth = 0;
+						App.Current.DrawingToolkit.DrawCircle (new Point (center.X + 1, center.Y + 1), radius);
```
vguzmanp: Should we add a property for the shadow offset? Or is it always 1?

SandraPG85: the shadow is usually one pixel, but if it is necessary when needed it can be added an offset and also a direction, left or right.

PR #516, opened 2018-01-08
Lon 1330

ylatuya:


VAS.Core/ViewModel/ProjectsManagerVM.cs
```diff
@@ -51,6 +51,7 @@ public ProjectsManagerVM ()
 				DescriptionText = ProjectsCreateHelper,
 				TipText = ProjectsCreateTip,
 			};
+			FilterText = "";
```
vmartos: this was already changed two lines after but intialized using string.Empty

vmartos: This was changed two lines after, but initialized using string.Empty;

VAS.Core/MVVMC/FastSmartWeakEvent.cs
```diff
@@ -0,0 +1,285 @@
...
+using System.Threading;
+
+namespace VAS.Core.MVVMC
+{
+	/// <summary>
+	/// A class for managing a weak event.
+	/// See http://www.codeproject.com/Articles/29922/Weak-Events-in-C
+	/// </summary>
+	/// <typeparam name="T">The delegate type of the event handlers.</typeparam>
+	public sealed class FastSmartWeakEvent<T> where T : class
```
vmartos: there is a new version of this in https://gist.github.com/dgrunwald/6445360 by the same author, I think is mostly the same.

VAS.Core/ViewModel/ProjectsManagerVM.cs
ylatuya: It looks like it has been introduced recently from someone else too, I will remove it

vmartos: yes it was fixed by me directly to master because it was crashing, Jorge needed this fix in order to change icons in ProjectManager (LongoMatch)

VAS.Core/MVVMC/FastSmartWeakEvent.cs
ylatuya: I believe it's the same one I used, why do you think it's a different one?

vmartos: well i see that you have this method:
https://github.com/fluendo/VAS/pull/516/files/7c6ce41b8bc46286be67c634bd53f238884d4e5a#diff-a4a9530773393b67ced78c6108dae07aR208
and the new version doesn't, maybe CanUseWeakEvent was added by you?

PR #521, opened 2018-01-10
LON-1056

dfernandezramos:


VAS.UI.Helpers.Gtk2/EntryHelper.cs
```diff
@@ -0,0 +1,49 @@
...
+	/// </summary>
+	public static class EntryHelper
+	{
+		/// <summary>
+		/// Autocomplete the specified entry with a given string list.
+		/// </summary>
+		/// <returns>The autocomplete.</returns>
+		/// <param name="list">List.</param>
+		/// <param name="entry">Entry.</param>
+		public static void Autocomplete (Entry entry, List<string> list)
```
ylatuya: you should use an extension method for that in ExtensionMethods instead

VAS.UI.Helpers.Gtk2/ExtensionMethods.cs
```diff
@@ -16,16 +16,38 @@
...
 using System;
+using System.Collections.Generic;
 using Gdk;
 using Gtk;
 
 namespace VAS.UI.Helpers
 {
 	/// <summary>
-	/// Provides extension methods for Gtk+
+	/// Autocomplete the specified entry with a given string list.
```
ylatuya: This documentation is for a particular method, not the whole class, it shouldn't be changed

PR #524, opened 2018-01-11
Bugfixes

ylatuya:


VAS.UI.Helpers.Gtk2/ImageView.cs
```diff
@@ -116,17 +116,15 @@ protected override void OnDestroyed ()
...
-					}
+				if (parent == null) {
+					return false;
+				}
+				if (!parent.Sensitive) {
+					return true;
 				}
-				return false;
+				// Handle correctly image in buttons
+				parent = parent.Parent?.Parent as Button;
```
vguzmanp: When does this happen? According to the getter's name, it should return true if `Parent.Sensitive`, but right now only does it if `Parent.Parent.Parent is Button && Parent.Parent.Parent.Sensitive`

VAS.UI.Helpers.Gtk2/Bindings/ButtonBinding.cs
```diff
@@ -53,10 +53,12 @@ protected override void UnbindView ()
 
 		protected override void UpdateView ()
 		{
-			button.Configure (image ?? Command.Icon,
-							  text ?? Command.Text,
-							  Command.ToolTipText, null);
-			button.Sensitive = Command.CanExecute ();
+			if (Command != null) {
```
vguzmanp: Squash this commit with the next one

VAS.Drawing/CanvasObjects/Timeline/TimeNodeView.cs
```diff
@@ -52,7 +52,6 @@ protected override void DisposeManagedResources ()
 				needle.Dispose ();
 				needle = null;
 			}
-			TimeNode = null;
```
vguzmanp: Without this the propertyChanged is still attached, is it intentional?

VAS.Drawing/CanvasObjects/Dashboard/TimerButtonView.cs
```diff
@@ -191,10 +191,16 @@ protected override void HandlePropertyChanged (object sender, PropertyChangedEve
 				return;
 			}
 
+			if (ViewModel.NeedsSync (e, nameof (ViewModel.TimerTime))) {
+				if (ViewModel.TimerTime == null) {
+					Active = false;
```
vguzmanp: Shouldn't it redraw too?

ylatuya: Changing Active already Redraw()

VAS.UI.Helpers.Gtk2/ImageView.cs
ylatuya: This is a special case in which an ImageView is inside a button and we want to draw the image insensitive when the Gtk.Button is insensitive. The Gtk.Button layouts the ImageView in a 3rd layer, so the Image is insensitive only if the it's 3rd parent is a button and it's not sensitive.

VAS.Drawing/CanvasObjects/Timeline/TimeNodeView.cs
ylatuya: There is an issue regarding how events are triggered, the first event removes the view and disposes it, but the canvas receives the event that something was removed after that and it needs to check against TimeNode which is now null, so the view is not removed from the Selection. In master I am fixing it in a different way

VAS.UI.Helpers.Gtk2/ImageView.cs
vguzmanp: Ok, so when does it happen that the ImageView has Parent but not Parent.Parent.Parent?

VAS.Drawing/CanvasObjects/Timeline/TimeNodeView.cs
vguzmanp: How are you fixing there?

VAS.UI.Helpers.Gtk2/ImageView.cs
ylatuya: I don't know exactly, it just happens as it's shown in the exception from hockey app, and this is the only NullReferenceException in the code.

VAS.Drawing/CanvasObjects/Timeline/TimeNodeView.cs
ylatuya: By emitting a event in the Canvas children each time a child removes internally a view, so that we remove it directly from the Selections instead of having to rely on other events like that fact that a period has been removed. I don't want to add this in release because it will require deeper changes that might be more intrusive

VAS.UI.Helpers.Gtk2/ImageView.cs
vguzmanp: 👌 but without a reproduction case, we can't test this properly :(

VAS.Drawing/CanvasObjects/Timeline/TimeNodeView.cs
vguzmanp: 👌

PR #526, opened 2018-01-11
Save previous filters on multiselection

VoldarGames:


VAS.Core/ViewModel/TimelineVM.cs
```diff
@@ -268,6 +268,8 @@ protected override void DisposeManagedResources ()
 		/// <value><c>true</c> if filters initialized; otherwise, <c>false</c>.</value>
 		public bool FiltersInitialized { get; set; }
 
+		public int SelectedProjects { get; set; }
```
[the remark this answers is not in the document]
VoldarGames: As we talked, I would use FiltersInitialized property

VAS.Services/Controller/EventsFilterController.cs
```diff
@@ -213,56 +224,34 @@ protected virtual void UpdateCommonTagsPredicates ()
 			}
 		}
 
-		bool GetPreviousActiveValue (string eventTypeName, bool defaultValue)
+		protected void UpdatePreviousPredicatesList (OrPredicate<TimelineEventVM> orPredicate)
```
vguzmanp: This could receive a `CompositePredicate<T>` as you don't really need it to be an Or

VAS.Services/Controller/EventsFilterController.cs
```diff
@@ -292,18 +281,23 @@ void AddAllEventTypePredicates (List<IPredicate<TimelineEventVM>> predicates, bo
 					composedPredicates.Add (new Predicate {
 						Name = Catalog.GetString ("No subcategories"),
 						Expression = eventTypeExpression.And (ev => !ev.Model.Tags.Intersect (tagList).Any ()),
-						Active = GetPreviousActiveValue (eventType.EventTypeVM.Name, defaultPredicateActiveValue)
+						Active = GetPreviousActiveValue (Catalog.GetString ("No subcategories"))
```
vguzmanp: This is a bit weird, the identifier shouldn't use the translation

VAS.Services/Controller/EventsFilterController.cs
```diff
@@ -43,7 +43,7 @@ namespace VAS.Services.Controller
 	/// </summary>
 	public abstract class EventsFilterController : ControllerBase<TimelineVM>
 	{
-		List<IPredicate<TimelineEventVM>> previousEventTypePredicateList = new List<IPredicate<TimelineEventVM>> ();
+		protected List<IPredicate<TimelineEventVM>> previousPredicateList = new List<IPredicate<TimelineEventVM>> ();
```
vguzmanp: You should update `TestEventsFilterController` with the previous predicate behaviour

VAS.Services/Controller/EventsFilterController.cs
VoldarGames: done as CompositePredicate<TimelineEventVM>, I can't use generic type on this method because we have previousPredicateList as field and this would convert this class to generic.

VAS.Services/Controller/EventsFilterController.cs
VoldarGames: As we talked there is one already on TestRATimelineVM

VAS.Tests/Services/TestEventsFilterController.cs
```diff
@@ -69,6 +72,29 @@ public void OneTimeSetUp ()
...
+			dummyEventsFilterController = new DummyEventsFilterController ();
+			dummyEventsFilterController.PreviousPredicates.Add (new Predicate ());
+			var onStartPreviousPredicatesCount = dummyEventsFilterController.PreviousPredicates.Count ();
+			//Act
+
+			try {
+				await dummyEventsFilterController.Stop ();
+				dummyEventsFilterController.ViewModel = null;
+
+			} catch (InvalidOperationException) {
```
vguzmanp: You shouldn't receive this Exception here, should you?

VoldarGames: This happens when you use ctrl-c ctrl-v 😄

PR #528, opened 2018-01-11
Bugfixes master

ylatuya:


VAS.Drawing/CanvasObjects/CanvasObject.cs
```diff
@@ -36,6 +37,7 @@ public abstract class CanvasObject : DisposableBase, ICanvasObject
 		protected CanvasObject ()
 		{
 			Visible = true;
+			Children = new RangeObservableCollection<ICanvasObject> ();
```
vguzmanp: With these changes, what's the difference between a Canvas and a CanvasObject?

I think it is clearer to have `Canvas` be the only thing that can contain other CanvasObjects, being itself a CanvasObject (Composite pattern). I understand that it's part of the idea (because of [your comment](https://github.com/fluendo/VAS/pull/528/commits/be79bcaf72c9965d156d18e6d568271917e6fa1c#diff-5d17e0c24f244d1633af34dd97d7f541R108) ) but I don't get how this fits that schema.

VAS.Drawing/CanvasObjects/CanvasObject.cs
```diff
@@ -82,16 +98,48 @@ protected CanvasObject ()
...
+
+		/// <summary>
+		/// Removes and object from the canvas.
+		/// </summary>
+		/// <param name="co">The object to remove.</param>
+		public void RemoveChild (ICanvasObject co)
+		{
+			co.RedrawEvent -= EmitRedrawEvent;
+			Children.Remove (co);
+			co.Dispose ();
```
vguzmanp: Here we are disposing something that we didn't create, is that right?

VAS.Drawing/CanvasObjects/CanvasObject.cs
```diff
@@ -105,18 +153,23 @@ public virtual void ClickReleased ()
...
 
-		protected void EmitClickEvent ()
+		public abstract void Draw (IDrawingToolkit tk, Area area);
+
+		protected void ClearChildren ()
 		{
-			if (ClickedEvent != null) {
-				ClickedEvent (this);
+			foreach (var child in Children.ToList ()) {
+				RemoveChild (child);
```
vguzmanp: This way the list will send out many deletion events

VAS.Drawing/CanvasObjects/CanvasObject.cs
```diff
@@ -185,6 +236,34 @@ public abstract class FixedSizeCanvasObject : CanvasObject
 
 	}
 
+	public abstract class CanvasSelectableObject : CanvasObject, ICanvasSelectableObject
```
vguzmanp: TODO: Review

VAS.Drawing/CanvasObjects/Timeline/TimelineView.cs
```diff
@@ -250,32 +204,15 @@ public override void Draw (IDrawingToolkit tk, Area area)
 			tk.End ();
 		}
 
-		public Selection GetSelection (Point point, double precision, bool inMotion = false)
+		public override Selection GetSelection (Point point, double precision, bool inMotion = false)
```
vguzmanp: TODO: Review

VAS.Drawing/CanvasObjects/Timeline/TimelineView.cs
```diff
@@ -56,15 +54,14 @@ public TimelineView ()
...
-			ClearObjects ();
 			selectionBorderL?.Dispose ();
 			selectionBorderR?.Dispose ();
 		}
 
 		public double SecondsPerPixel {
 			set {
 				secondsPerPixel = value;
-				foreach (TimeNodeView to in nodes) {
+				foreach (TimeNodeView to in Children) {
```
vguzmanp: Now children can contain any `ICanvasObject`

VAS.Drawing/CanvasObjects/Timeline/TimelineView.cs
```diff
@@ -225,7 +179,7 @@ public override void Draw (IDrawingToolkit tk, Area area)
 
 			tk.Begin ();
 			DrawBackground (tk, area);
-			foreach (TimeNodeView p in nodes) {
+			foreach (TimeNodeView p in Children) {
```
vguzmanp: Now children can contain any `ICanvasObject`

VAS.Drawing/SelectionCanvas.cs
```diff
@@ -332,6 +333,16 @@ protected virtual Selection GetSelection (Point coords, bool inMotion = false, b
 			return sel;
 		}
 
+		protected override void HandleChildrenChanged (object sender, NotifyCollectionChangedEventArgs e)
+		{
+			if (e.Action == NotifyCollectionChangedAction.Remove) {
+				var removedViews = e.OldItems.OfType<ICanvasObject> ().ToList ();
```
vguzmanp: You don't need the `ToList`

VAS.Drawing/CanvasObjects/CanvasObject.cs
ylatuya: I think your are not understanding these changes. Think of the hierarchy as the one in Gtk+ where a widget (a CanvasObject) has as children other widget (CanvasObject) and the toplevel Window (Canvas) is also a Widget.
Here I am just moving the logic in the TimelineView to the CanvasObject base class that any CanvasObject can include other CanvasObject too like the TimelineView was doing

VAS.Drawing/CanvasObjects/CanvasObject.cs
```diff
@@ -82,16 +98,48 @@ protected CanvasObject ()
...
+
+		/// <summary>
+		/// Inserts a child at a given position.
+		/// </summary>
+		/// <param name="index">Index</param>
+		/// <param name="child">The child to add</param>
+		protected void InsertChild (int index, ICanvasObject child)
+		{
+			Children.Insert (index, child);
+			child.RedrawEvent += EmitRedrawEvent;
```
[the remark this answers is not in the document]
ylatuya: This only forward DrawEvent's, the event has an area that is specific to the child

VAS.Drawing/CanvasObjects/CanvasObject.cs
vguzmanp: Following your Gtk+ example, the only things with Children are `GtkContainer` objects, and subclasses of it, Labels are widgets but don't have Children.

If Children is in Canvas, we have the exact same behaviour: you can either inherit from CanvasObject and NOT have Children (Gtk.Label), or inherit from Canvas, and have Children (Gtk.Box).

I think having the list in what will be the base class will make it more difficult to maintain, as all objects will have lists of children, even when unused; and the Canvas object is kind of useless, as I could use a CanvasObject directly.

VAS.Drawing/SelectionCanvas.cs
ylatuya: It's needed because en IEnumerable does not have Contains

VAS.Drawing/CanvasObjects/Timeline/TimelineView.cs
ylatuya: Because a TimelineView is CanvasObject, the Children property is exposed in the base class

VAS.Drawing/CanvasContainer.cs
```diff
@@ -0,0 +1,157 @@
...
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Linq;
+using VAS.Core.Common;
+using VAS.Core.Interfaces.Drawing;
+using VAS.Core.Store.Drawables;
+
+namespace VAS.Drawing
+{
+	public class CanvasContainer : CanvasObject, ICanvasSelectableObject, INotifyCollectionChanged,
```
vguzmanp: Missing class documentation

VAS.Drawing/CanvasContainer.cs
```diff
@@ -0,0 +1,157 @@
...
+				if (selection == null || tmp.Accuracy < selection.Accuracy) {
+					selection = tmp;
+				}
+			}
+			return selection;
+		}
+
+		public virtual void Move (Selection s, Point dst, Point start)
+		{
+			throw new NotImplementedException ("Move not supported for this canvas container");
```
vguzmanp: You could add `GetType()` to the exception message, so that we get the subclass name if this is called.

VAS.Drawing/CanvasContainer.cs
```diff
@@ -0,0 +1,157 @@
...
+		protected override void DisposeManagedResources ()
+		{
+			Clear ();
+		}
+
+		public int Count => children.Count;
+
+		public bool IsReadOnly => true;
+
+		public IEnumerator<ICanvasObject> GetEnumerator ()
```
vguzmanp: I think you can do the same structure with `=>` in methods as it's done in properties

VAS.Drawing/CanvasContainer.cs
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		public ICanvasObject this [int index] {
+			get { return children [index]; }
+		}
+
+		public void Clear ()
+		{
+			foreach (var child in children) {
+				RemoveChild (child, false);
```
vguzmanp: This will send out several collectionChanged events

VAS.Tests/Drawing/Widgets/TestPlaysTimeline.cs
```diff
@@ -163,7 +163,7 @@ public void RemovePeriodNode_PeriodNodeSelected_SelectionUpdated ()
 			var timeNode = new TimeNode { Start = new Time (0), Stop = new Time (2) };
 			project.Timers [0].Nodes.Add (timeNode);
 			var timerTimeline = timeline.Objects [0] as TimerTimelineView;
-			timeline.UpdateSelection (new Selection (timerTimeline.nodes [0], SelectionPosition.All, 0));
+			timeline.UpdateSelection (new Selection (timerTimeline [0] as IMovableObject, SelectionPosition.All, 0));
```
vguzmanp: Why the cast?

VAS.Drawing/CanvasObjects/Timeline/TimelineView.cs
```diff
@@ -56,15 +54,14 @@ public TimelineView ()
...
-			ClearObjects ();
 			selectionBorderL?.Dispose ();
 			selectionBorderR?.Dispose ();
 		}
 
 		public double SecondsPerPixel {
 			set {
 				secondsPerPixel = value;
-				foreach (TimeNodeView to in nodes) {
+				foreach (TimeNodeView to in this) {
```
vguzmanp: In the previous version I commented:

> Now children can contain any ICanvasObject

The thing is that with this line, if anything is not a `TimeNodeView` it will throw an exception

ylatuya: Yes but a TimelineView only has TimeNodeView as children

VAS.Drawing/CanvasContainer.cs
ylatuya: Nope :) I changed it based on your previous review so that passing false do not remove the element from the collection. The only event is sent calling Clear() after this foreach

VAS.Drawing/CanvasContainer.cs
ylatuya: done

VAS.Drawing/CanvasContainer.cs
ylatuya: done

VAS.Drawing/CanvasContainer.cs
ylatuya: done

VAS.Drawing/CanvasObjects/Timeline/TimelineView.cs
ylatuya: I can change it too

VAS.Drawing/CanvasContainer.cs
vguzmanp: nice! 👌

PR #531, opened 2018-01-16
Fix ProjectVM Edited property summary

axelgorris:


VAS.Services/Controller/ProjectsController.cs
```diff
@@ -163,7 +163,7 @@ protected virtual Task HandleNew (CreateEvent<TModel> evt)
 			TViewModel selectedVM = ViewModel.Selection.FirstOrDefault ();
 
 			if (selectedVM != null) {
-				if (ViewModel.LoadedProject.Model != null && ViewModel.LoadedProject.IsChanged) {
+				if (ViewModel.LoadedProject.Model != null && ViewModel.LoadedProject.Edited) {
 					await Save (ViewModel.LoadedProject, false);
 				}
```
SandraPG85: add a test changing the playing to ensure that model is checked instead of the viewmodel

PR #535, opened 2018-01-22
Lon 1343

axelgorris:
#### Things to review:
- [ ] Unit tests
- [ ] Sorted usings

### DIFF
--- a/VAS.Services/VideoPlayerController.cs
+++ b/VAS.Services/VideoPlayerController.cs
@@ -639,6 +639,7 @@ namespace VAS.Services
 		{
 			UpdatePlayingState (false);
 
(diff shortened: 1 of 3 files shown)


VAS.Services/VideoPlayerController.cs
```diff
@@ -729,6 +730,8 @@ public virtual void UnloadCurrentEvent ()
 			Reset ();
 			loadedEvent = null;
 			LoadedTimelineEvent = null;
+			PlayerVM.FrameDrawing = null;
```
ylatuya: Can you also add a unit test for that or extend the existing one to check that FrameDrawing is correctly reset when closing the current event

PR #544, opened 2018-01-25
Add FillCanvas

VoldarGames:


VAS.Drawing/FillCanvas.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace VAS.Drawing
+{
+	/// <summary>
+	/// Canvas that for each object of type <see cref="FixedSizeCanvasObject"/>
+	/// sets its size to this canvas associated widget's size
+	/// </summary>
+	public class FillCanvas : Canvas
+	{
+		public FillCanvas (IWidget widget) : base (widget)
```
ylatuya: Missing unit tests for this canvas

VAS.Drawing/FillCanvas.cs
```diff
@@ -0,0 +1,45 @@
...
+//
+using System;
+using System.Linq;
+using VAS.Core.Common;
+using VAS.Core.Interfaces.Drawing;
+
+namespace VAS.Drawing
+{
+	/// <summary>
+	/// Canvas that for each object of type <see cref="FixedSizeCanvasObject"/>
```
ylatuya: I would change the description to
A Canvas that layouts its children to fill and expand to the size of the canvas. It should be used with a single child that will be allocated to take all the space of the widget.

PR #548, opened 2018-01-30
Create a method to copy entire directories

dfernandezramos:
#### Things to review:
- [ ] Unit tests
- [ ] Sorted usings

### DIFF
--- a/VAS.DB/CouchbaseStorage.cs
+++ b/VAS.DB/CouchbaseStorage.cs
@@ -42,6 +42,7 @@ namespace VAS.DB
 		string storageName;
 		object mutex;
 		bool documentUpdated;
+		string dbDir;
 
 		public CouchbaseStorage (Database db)
 		{
@@ -59,9 +60,8 @@ namespace VAS.DB
 		public CouchbaseStorage (string dbDir, string storageName)
 		{
 			this.storageName = storageName;
-			var manager = (CouchbaseManager)App.Current.DependencyRegistry.Retrieve<IStorageManager> (InstanceType.New, dbDir);
-			db = manager.OpenDatabase (storageName);
-			Init ();
+			this.dbDir = dbDir;
+			Start ();
 		}
 
 		void Init ()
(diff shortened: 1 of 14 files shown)


VAS.Core/Common/Utils.cs
```diff
@@ -290,6 +290,43 @@ public static DateTime GetNetworkTime ()
 			return networkDateTime;
 		}
 
+		/// <summary>
+		/// Copy a whole directory to the specified.
+		/// </summary>
+		/// <param name="sourceDirName">Source dir name.</param>
+		/// <param name="destDirName">Destination dir name.</param>
+		/// <param name="copySubDirs">If set to <c>true</c> copy sub dirs.</param>
+		public static void DirectoryCopy (string sourceDirName, string destDirName, bool copySubDirs)
```
axelgorris: Should the name be Path instead of name? It may be misleading

VAS.DB/CouchbaseStorage.cs
```diff
@@ -105,6 +107,16 @@ void Init ()
 			}
 		}
 
+		/// <summary>
+		/// Retrieves a new instance of the current storage and initializes it again.
+		/// </summary>
+		public void Restart ()
+		{
+			var manager = (CouchbaseManager)App.Current.DependencyRegistry.Retrieve<IStorageManager> (InstanceType.New, dbDir);
```
ylatuya: You should close the old one.
Reload or Reopen might make more sense than restart.
Call this mehod on the constructor instead of copy pasting it

VAS.Core/Common/Utils.cs
```diff
@@ -290,6 +290,43 @@ public static DateTime GetNetworkTime ()
 			return networkDateTime;
 		}
 
+		/// <summary>
+		/// Copy a whole directory to the specified.
```
ylatuya: incomplete sentence

VAS.Core/Common/Utils.cs
ylatuya: Missing unit tests
I would rename it to CopyDirectory wich is more natural and use sourcePath and destPath since name indicates a name not a path

PR #549, opened 2018-01-30
Add the option to create a PlaylistElementVM based on a TimelineEventVM

vmartos:
- This fixes a bug where the eye was not visible due to creation of other TimelineEventVM Instance

#### Things to review:
- [x] Unit tests
- [x] No hardcoded resources (strings, sizes, colors...)
- [x] Sorted usings

### DIFF
--- a/VAS.Core/Interfaces/IViewModelFactoryService.cs
+++ b/VAS.Core/Interfaces/IViewModelFactoryService.cs
@@ -1,0 +1,37 @@
+﻿//
+//  Copyright (C) 2018 Fluendo S.A.
+//
+//  This program is free software; you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation; either version 2 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+using System;
+using VAS.Core.ViewModel;
+using VAS.Core.Store;
+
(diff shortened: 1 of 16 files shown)


VAS.Core/ViewModel/PlaylistElementVM.cs
```diff
@@ -73,7 +73,7 @@ public class PlaylistElementVM : ViewModelBase<IPlaylistElement>, IPlayable
 		/// Gets or sets a value indicating whether this <see cref="T:VAS.Core.ViewModel.PlaylistElementVM"/> is playing.
 		/// </summary>
 		/// <value><c>true</c> if playing; otherwise, <c>false</c>.</value>
-		public bool Playing { get; set; }
+		public virtual bool Playing { get; set; }
```
VoldarGames: Why this needs to be virtual, instead of using a simple property?

VoldarGames: in case it is needed, please summarize it

vmartos: because PlaylistPlayElementVM needs to override this property to change the TimelineEventVM Playing property. Because the TreeViews are using TimelineEventVM's to render cells, so for PlaylistPlayElementVM we need to use the TimelineEventVM.Playing property, not for the rest of PlaylistElementVM

VoldarGames: 👍 can you summarize it?

VAS.Core/Interfaces/IViewModelFactoryService.cs
```diff
@@ -0,0 +1,47 @@
...
+	/// An interface for creating correctly typed ViewModel instances based on it's model, this is useful to work with
+	/// base classes and create child viewmodels without knowing the reference on it.
+	/// </summary>
+	public interface IViewModelFactoryService
+	{
+		/// <summary>
+		/// Gets the TypeMappings Dictionary, where Key is the Model Type and Value is the ViewModel Type
+		/// </summary>
+		/// <value>The Type Mappings</value>
+		Dictionary<Type, Type> TypeMappings { get; }
```
ylatuya: Is this required to be public now?
Returning a dicctionary any one can mess with it and you don't really know how to use it. Instead you can have a public method AddMapping that internally uses that diccionary.

PR #555, opened 2018-02-02
Lon 1376

vguzmanp:
#### Things to review:
- [ ] Unit tests
- [ ] Sorted usings


VAS.Core/MVVMC/ControllerBase.cs
```diff
@@ -48,6 +48,15 @@ protected override void DisposeUnmanagedResources ()
 
 		#region IController
 
+		public bool Started {
```
ylatuya: convert ti to auto-properties

VAS.Core/Interfaces/MVVMC/IViewLocator.cs
```diff
@@ -0,0 +1,47 @@
...
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+//
+
+using System;
+
+namespace VAS.Core.Interfaces.MVVMC
+{
+	public interface IViewLocator
```
ylatuya: This interface is the same for the ControllerLocator probably

ylatuya: Maybe ViewLocator should be a Registry, there is no point on having a particular implementation for conmtrollers and views

VAS.Core/MVVMC/LocatorBase.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Linq;
+using System.Collections.Generic;
+using VAS.Core.Interfaces.MVVMC;
+
+namespace VAS.Core.MVVMC
+{
+	/// <summary>
+	/// A location service for <see cref="TReturn"/> components
+	/// </summary>
+	public class LocatorBase<TReturn> : ILocator<TReturn>
```
ylatuya: I don't think this should have Base in the name, it's not a base class but a complete implementation of ILocator

VAS.Core/MVVMC/ViewLocator.cs
```diff
@@ -24,53 +24,7 @@ namespace VAS.Core.MVVMC
 	/// <summary>
 	/// A location service for IView components.
 	/// </summary>
-	public class ViewLocator
+	public class ViewLocator : LocatorBase<IView>
```
ylatuya: Not sure if we want a file for a generic, in other place  generic specialization are in the same file as the generic, since they are just a more friendly way if naming the class but they don't add value

VAS.Core/App.cs
```diff
@@ -36,8 +37,8 @@ public abstract class App
 		public IStorageManager DatabaseManager;
 		public IKpiService KPIService;
 		public IJobsManager JobsManager;
-		public ViewLocator ViewLocator;
-		public ControllerLocator ControllerLocator;
+		public ILocator<IView> ViewLocator;
```
ylatuya: You still have the class ViewLocator: ILocator\<IView\>
The purpose of this class is to use it as a shortcut for ILocator\<IView\>, so or you use it instead of using ILocator\<View\> or you remove it

VAS.UI.Gtk2/UI/Panel/PreferencesPanel.cs
```diff
@@ -31,7 +31,7 @@
 namespace VAS.UI.Panel
 {
 	[System.ComponentModel.ToolboxItem (true)]
-	[ViewAttribute (PreferencesState.NAME)]
+	[ViewAttribute (PreferencesState.NAME, 0)]
```
ylatuya: 0 is already the default value

vguzmanp: I liked it being explicit here, as it's the only one that gets overwritten right now.

PR #557, opened 2018-02-02
Simplify makefiles and remove autotools

ylatuya:
#### Things to review:
- [ ] Unit tests
- [ ] Sorted usings


build/genmakefile.config
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+echo '🦖 🦖 🦖 Generating makefile ...'
```
vmartos: longirex? :D

tools/resize_svg.py
```diff
@@ -68,12 +70,18 @@ def main(root=None):
     if not root:
         root = '.'
     di = os.path.join(root, 'data')
+    md5, currentmd5 = checksum.read_checksum(di)
+    if md5 == currentmd5:
```
vmartos: can we have a helper function that gives if the checksum has changed? returns true or false, instead of comparing it later

tools/runtools_vas.py
```diff
@@ -14,10 +14,10 @@
 
 application_tools = [runtools]
 # update_data should be the last one executed, as it depends on everything being already generated
-desktop_tools = [resize_svg, update_data]
```
vmartos: why update_data script is removed? I mean this updates Makefile.am, and now we don't need them, but we need somewhere to list the files to be installed or it is hardcoded as copy data to the prefix/share?

ylatuya: The script was basically listing files in git to fill the Makefile.am, now we do the opposite and install these files directly.

tools/resize_svg.py
ylatuya: I though to do it that way but finally I decided to do it this way just in case  someone needed the actual value.
I can change it if you want

tools/runtools_vas.py
vmartos: 👍

PR #560, opened 2018-02-07
Extract the parseModifier to a public method

vguzmanp:
It can be used with EventKeys and EventButtons, although they don't have any common interface with the 'State' property

#### Things to review:
- [ ] Unit tests
- [ ] Sorted usings

### DIFF
--- a/VAS.Core/Common/Keyboard.cs
+++ b/VAS.Core/Common/Keyboard.cs
@@ -15,9 +15,10 @@
 //  along with this program; if not, write to the Free Software
 //  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 //
+using System;
 using Gdk;
-using VAS.Core.Store;
 using VAS.Core.Interfaces;
+using VAS.Core.Store;
 
 namespace VAS.Core.Common
 {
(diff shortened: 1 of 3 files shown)


VAS.Core/Common/Keyboard.cs
```diff
@@ -35,37 +36,25 @@ public string NameFromKeyval (uint keyval)
 
 		public HotKey ParseEvent (object obj)
 		{
-			int modifier = 0;
-			EventKey evt = obj as EventKey;
+			EventKey eventKey = obj as EventKey;
 
-			if (IsSupportedModifier (evt.Key)) {
+			int keyval = 0;
```
ylatuya: Why is keyval intialized here now?

VAS.Core/Common/Keyboard.cs
```diff
@@ -74,6 +63,27 @@ public HotKey ParseEvent (object obj)
 			};
 		}
 
+		public int ParseModifier (object originalEvent)
```
ylatuya: Why is this function public but not added to the interface?
 Why can't you use ParseEvent and keep the interface simplier?

VAS.Core/Common/Keyboard.cs
```diff
@@ -187,4 +197,32 @@ bool IsSupportedModifier (Key key)
...
+	/// This class wraps over both EventKey and EventButton and exposes their State property.
+	/// </summary>
+	class EventWrapper
+	{
+		EventKey eventKey;
+		Gdk.EventButton eventButton;
+
+		public EventWrapper (object evnt)
+		{
+			EventKey key = evnt as EventKey;
```
ylatuya: Why do you need these local variables at all ?
eventKey =  evnt as Gdk.EventKey;
eventButton = event as Gdk.EventButton;
if (eventButton != null && eventKey != null) throw new

VAS.Core/Common/Keyboard.cs
vguzmanp: Oops

VAS.Core/Common/Keyboard.cs
vguzmanp: It exposes a gtk-only way of doing things. Other platforms don't necessarily use an int to store modifier flags, so I thought it wouldn't make sense in the interface.

VAS.Core/Common/Keyboard.cs
vguzmanp: Ok

VAS.Core/Common/Keyboard.cs
ylatuya: It exposes the modifiers which are not gtk specific, under the HotkKey.Modifier property.
You are creating a new API when you can use an existing one:
int modifier = KeyBoard.ParseEvent(event).Modifier

vguzmanp: Right, I'll move it to the interface.

Calling ParseEvent has the problem that HotKeys have a `Key` property that doesn't make sense here, as it's a mouse press, and I think that Hotkey is good as it is, without having extra logic to manage mouse/keyboard.

ylatuya: ok!

PR #563, opened 2018-02-09
Lon 1373

ylatuya:
#### Things to review:
- [ ] Unit tests
- [x] No hardcoded resources (strings, sizes, colors...)
- [x] Sorted usings
- [x] Add summaries where needed 

### DIFF
--- a/VAS.Multimedia/Utils/Devices.cs
+++ b/VAS.Multimedia/Utils/Devices.cs
@@ -40,15 +40,15 @@ namespace VAS.Multimedia.Utils
 
 		[DllImport ("libvas.dll")]
 		static extern IntPtr lgm_device_video_format_get_info (IntPtr raw, out int width, out int height,
-		                                                       out int fps_n, out int fps_d);
+															   out int fps_n, out int fps_d);
 
-		static readonly string[] devices_osx = new string[2] { "avfvideosrc", "decklinkvideosrc" };
-		static readonly string[] devices_win = new string[2] { "ksvideosrc", "dshowvideosrc" };
-		static readonly string[] devices_lin = new string[2] { "v4l2src", "dv1394src" };
+		static readonly string [] devices_osx = { "avfvideosrc", "decklinkvideosrc" };
+		static readonly string [] devices_win = { "ksvideosrc", "dshowvideosrc" };
+		static readonly string [] devices_lin = { "v4l2src", "dv1394src" };
 
 		static public List<Device> ListVideoDevices ()
 		{
-			string[] devices;
+			string [] devices;
 			if (VAS.Core.Common.Utils.OS == OperatingSystemID.OSX)
 				devices = devices_osx;
 			else if (VAS.Core.Common.Utils.OS == OperatingSystemID.Windows)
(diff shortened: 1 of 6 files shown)


VAS.Multimedia/Utils/Devices.cs
```diff
@@ -76,6 +78,9 @@ static public List<Device> ListVideoDevices ()
 					device.DeviceType = CaptureSourceType.System;
 					device.SourceElement = source;
 					device.ID = deviceName;
+					if (source == "gdiscreencapsrc" || source == "dx9screencapsrc") {
+						device.Prefix = Monitor;
```
vguzmanp: Why is only the windows capturer the one with `Monitor`?

VAS.Multimedia/Utils/Devices.cs
```diff
@@ -76,6 +78,9 @@ static public List<Device> ListVideoDevices ()
 					device.DeviceType = CaptureSourceType.System;
 					device.SourceElement = source;
 					device.ID = deviceName;
+					if (source == "gdiscreencapsrc" || source == "dx9screencapsrc") {
```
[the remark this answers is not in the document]
ylatuya: I can't use an enum, we need the soure string, I can use constants though

libvas/lgm-device.c
```diff
@@ -362,3 +350,32 @@ lgm_device_enum_audio_devices (const gchar * device)
...
+  else if (!g_strcmp0 (source_name, "avfvideosrc"))
+    prop_name = "device";
+  else if (!g_strcmp0 (source_name, "osxscreencapsrc"))
+    prop_name = "device";
+  else if (!g_strcmp0 (source_name, "decklinkvideosrc"))
+    prop_name = "device-number";
+  else if (!g_strcmp0 (source_name, "dx9screencapsrc"))
+    prop_name = "monitor";
+  else if (!g_strcmp0 (source_name, "gdiscreencapsrc"))
+    prop_name = "monitor";
```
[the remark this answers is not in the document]
ylatuya: that's the name of the property used by the gstreamer to set the device id, in these new ones it's called "monitor"

VAS.Multimedia/Utils/Devices.cs
ylatuya: these devices don't list devices with their name, just by their index, the prefix helps to convert "1" into "Monitor 1" which is what the user will see

VAS.Multimedia/Utils/Devices.cs
```diff
@@ -19,8 +19,10 @@
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
+using VAS.Core;
 using VAS.Core.Common;
-using VAS.Multimedia.Capturer;
+using static VAS.Core.Resources.Strings;
```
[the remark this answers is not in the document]
ylatuya: We will have to discuss that to standardize because I think it's a good use of static.

VAS.Multimedia/Utils/Devices.cs
vguzmanp: 👌 but why only add it to the windows ones? How will the osx one be shown?

ylatuya: The OSX ones will be shown as the source element returns them, for example LCD Screen, Samsung TV, Blackmagic 2, Facetime HD

PR #568, opened 2018-02-13
Fix Contract Resolvers to use always same instance for cache

vmartos:
- Divide IsChangedContractResolver into two resolvers:
    - IsChangedContractResolver and ClonerContractResolver
- Improves performance speed for Cloning objects by reusing the
same contract resolver instance that caches types
- Also fix how new playlist elements are created based on TimelineEvent,
Basically we need to clone the timeline event to fix corrupted database

#### Things to review:
- [ ] Unit tests
- [ ] Sorted usings

### DIFF
--- a/VAS.Core/Serialization/Serializer.cs
+++ b/VAS.Core/Serialization/Serializer.cs
@@ -116,7 +116,7 @@ namespace VAS.Core.Serialization
 			case SerializationType.Json:
 				StreamReader sr = new StreamReader (stream, Encoding.UTF8);
 				JsonSerializerSettings settings = JsonSettings;
-				settings.ContractResolver = new IsChangedContractResolver ();
+				settings.ContractResolver = IsChangedContractResolver.Instance;
 				return JsonConvert.DeserializeObject (sr.ReadToEnd (), type, settings);
 			default:
 				throw new Exception ();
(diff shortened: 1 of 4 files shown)


VAS.Core/Serialization/Serializer.cs
```diff
@@ -315,40 +315,46 @@ int GetModifierValue (string serializedValue)
 		}
 	}
 
-	public class IsChangedContractResolver : DefaultContractResolver
+	/// <summary>
+	/// Cloner contract resolver, it ignores all properties and fields with NonSerialized and CloneIgnore attributes
```
ylatuya: The description is a little bit confusing, it whould be:
Serializes all writable properties, even those marked as JsonIgnore and only ingnore the properties marked as CloneIgnore or NonSerialized

VAS.Core/Serialization/Serializer.cs
```diff
@@ -315,40 +315,46 @@ int GetModifierValue (string serializedValue)
...
+	/// it does not ignore any properties, even if it has the JsonIgnore Attribute
+	/// </summary>
+	public class ClonerContractResolver : DefaultContractResolver
 	{
-		public bool IgnoreJsonIgnore { get; private set; }
+		public static readonly ClonerContractResolver Instance = new ClonerContractResolver ();
```
ylatuya: It should follow our Singleton implementation guidelines

VAS.Core/Serialization/Serializer.cs
```diff
@@ -315,40 +315,46 @@ int GetModifierValue (string serializedValue)
...
 		}
+	}
 
-		protected override JsonContract CreateContract (Type type)
+	/// <summary>
+	/// Is Changed Contract resolver, it sets the IsChanged property to false, used mainly for deserialize/Load
+	/// </summary>
+	public class IsChangedContractResolver : DefaultContractResolver
+	{
+		public static readonly IsChangedContractResolver Instance = new IsChangedContractResolver ();
```
ylatuya: It should follow our Singleton implementation guidelines

